PR #1, opened 2020-06-01
Project structure proposal

jccr:
Formatted version [here](https://github.com/jccr/web/blob/project-structure-proposal/org/proposals/001-project-structure.md).


jccr:
Updated with explanations on tooling proposals.

PR #3, opened 2020-07-10
Shared models

JayPanoz:
This PR inits shared models for Web (using TypeScript), based on the latest architecture docs/proposals/APIs. 

One significant note is that we don’t have unit tests yet.


shared-models/src/Format/Format.ts
```diff
@@ -0,0 +1,220 @@
+/* Copyright 2020 Readium Foundation. All rights reserved.
+ * Use of this source code is governed by a BSD-style license,
+ * available in the LICENSE file present in the Github repository of the project.
+ */
+
+import { MediaType } from "./MediaType";
+
+/** Represents a known file format, uniquely identified by a media type. */
+export default class Format {
```
mickael-menu, line 9: Is `Format` implicitly equatable? We will need to do things like:

```typescript
format == Format.epub()
```

(Also, if there's a way to have `format == Format.epub`, it would be nicer)

mickael-menu: These comments apply to `MediaType` as well.

shared-models/src/Publication/Link.ts
```diff
@@ -0,0 +1,222 @@
...
+  public templated?: boolean;
+  
+  /** MIME type of the linked resource. */
+  public type?: string;
+  
+  /** Title of the linked resource. */
+  public title?: string;
+  
+  /** Relation between the linked resource and its containing collection. */
+  public rels?: Array<string>;
```
mickael-menu: `rels` could be more convenient/safer as a `Set` instead of `Array`.

It's something I will probably change on Swift, but Kotlin is using a `Set` at the moment.

shared-models/src/Publication/Link.ts
```diff
@@ -0,0 +1,222 @@
...
+      return new Set();
+    } else {
+      return new URITemplate(this.href).parameters;
+    }
+  }
+
+  /** Expands the `Link`'s HREF by replacing URI template variables by the given parameters.
+   *  See RFC 6570 on URI template: https://tools.ietf.org/html/rfc6570
+   */
+  public expandTemplate(parameters: { [param: string]: string}): Link {
```
mickael-menu, line 118: I'm a bit confused about the type signature, does it allow to do something like this?

```typescript
searchLink.expand({query: "tomatoe"})
```

shared-models/src/Publication/Encryption/Properties+Encryption.ts
```diff
@@ -0,0 +1,17 @@
...
+import { IEncryption } from "./Encryption";
+
+declare module "../Properties" {
+  export interface Properties {
+    getEncryption: () => IEncryption | null;
+  }
+}
+
+Properties.prototype.getEncryption = function() {
+  return this.otherProperties["encryption"] || null;
```
mickael-menu, line 16: You probably need to parse the JSON to return a proper `Encryption` object here, otherwise the caller will get a raw JSON array.

https://github.com/readium/r2-shared-swift/blob/f096f4c84c10e83f887217f05a4d203379ab2645/r2-shared-swift/Publication/Encryption/Properties%2BEncryption.swift#L18-L27

shared-models/src/Publication/Metadata.ts
```diff
@@ -0,0 +1,135 @@
...
+  public narrator: Array<IContributor>;
+  public penciler: Array<IContributor>;
+  public translator: Array<IContributor>;
+  public languages: Array<string>;
+  public description?: string;
+  public publisher: Array<IContributor>;
+  public imprint: Array<IContributor>;
+  public published: Date | null;
+  public modified: Date | null;
+  public subject: Array<Subject>;
```
mickael-menu: Using noun plurals when it's returning an `Array` would be nicer, IMHO.

shared-models/src/Publication/Metadata.ts
```diff
@@ -0,0 +1,135 @@
...
+  public letterer: Array<IContributor>;
+  public narrator: Array<IContributor>;
+  public penciler: Array<IContributor>;
+  public translator: Array<IContributor>;
+  public languages: Array<string>;
+  public description?: string;
+  public publisher: Array<IContributor>;
+  public imprint: Array<IContributor>;
+  public published: Date | null;
+  public modified: Date | null;
```
mickael-menu, line 65: What's the difference between

```typescript
public published?: Date
// and
public published: Date | null
```

If it's the same, maybe we should pick one to be consistent.

shared-models/src/Publication/Metadata.ts
```diff
@@ -0,0 +1,135 @@
...
+  public belongsToSeries?: Array<Collection>;
+
+  /** This contains the reading progression as declared in the manifest, so it might be 
+   *  `auto`. To know the effective reading progression used to lay out the content, use
+   *  `effectiveReadingProgression` instead.
+  */
+  public readingProgression: ReadingProgression;
+  public duration: number | null;
+  public numberOfPages: number | null;
+  public otherMetadata: JSONDictionary;
```
mickael-menu: Exposing publicly the `JSONDictionary` means that the reading app could modify its contents. Instead we could have something like:

```typescript
private _otherMetadata: JSONDictionary;

public otherMetadata(): [key: string] : string {
    return _otherMetadata.json
}
```

By the way I introduced [`JSONDictionary`](https://github.com/readium/r2-shared-swift/blob/develop/r2-shared-swift/Publication/Publication%2BJSON.swift#L30) on Swift for two purposes:

* Make the models equatable automatically, because using `[String: Any]` would mean writing a custom `isEqual` function since `Any` is not equatable, which was not convenient.
* Easily pop properties from it.

But you might not need it in Typescript, for example [in Kotlin we use a `Map<String, Any>` directly](https://github.com/readium/r2-shared-kotlin/blob/develop/r2-shared/src/main/java/org/readium/r2/shared/publication/Metadata.kt#L65).

shared-models/src/Publication/Metadata.ts
```diff
@@ -0,0 +1,135 @@
...
+    this.published = json.parseDate("published");
+    this.modified = json.parseDate("modified");
+    this.subject = json.parseArray("subject");
+    const belongsTo = json.parseRaw("belongsTo");
+    this.belongsToCollection = belongsTo ? belongsTo["collection"] : [];
+    this.belongsToSeries = belongsTo ? belongsTo["series"] : [];
+    this.readingProgression = json.parseRaw("readingProgression") || ReadingProgression.auto;
+    this.duration = json.parsePositive("duration");
+    this.numberOfPages = json.parsePositive("numberOfPages");
+    this.otherMetadata = json;
```
mickael-menu: Here `otherMetadata` will contain all the JSON of the manifest, instead of only the extra ones which don't fit in the RWPM spec. To avoid this, we need to "pop" the properties when parsing them. Notice the `json.pop()` in Swift: https://github.com/readium/r2-shared-swift/blob/f096f4c84c10e83f887217f05a4d203379ab2645/r2-shared-swift/Publication/Metadata.swift#L102-L130

shared-models/src/Publication/Publication.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /** The URL where this publication is served, computed from the `Link` with `self` relation.
+   *  e.g. https://provider.com/pub1293/manifest.json gives https://provider.com/pub1293/
+   */
+  public baseURL(): string {
+    const selfLink = this.manifest.links.find(el => el.rels.includes("self"));
+    if (selfLink) {
+      return selfLink.href;
+    } else {
+      return window.location.href;
```
mickael-menu: I'm not sure about returning `window.location.href` here, but it might be some special Web concerns that I'm not aware of.

The other possibility would be to return `null` and let the caller (navigator probably?) handle a fallback on what's appropriate.

shared-models/src/Publication/divina/Publication+DiViNa.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    /** Provides navigation to positions in the Publication content that correspond to the locations
+     *  of page boundaries present in a print source being represented by this EPUB Publication.
+     */
+    getGuided: () => Links;
+  }
+}
+
+Publication.prototype.getGuided = function() {
+  return findValue(this.subcollections, "guided") || new Links([]);
```
mickael-menu: If I'm not missing something, my guess is that the `findValue()` will return a `CoreCollection`, maybe a `?.links` is missing?

shared-models/src/Format/Format.ts
JayPanoz: For objects, “JS equality” checks that the two objects occupy the same place in memory, it doesn’t check that both objects have the same property keys and values – [this article can serve as an explanation of the issue.](https://dev.to/joshuapbritz/why-it-s-so-hard-to-check-object-equality-in-javascript-8a0)

So it’s kinda always been up to devs to pick their preferred solution for that i.e. `json.stringify()`, shallow/deep equality – very often by using a library such as lodash, underscore, etc. so it can be very opinionated.

shared-models/src/Publication/Link.ts
JayPanoz: Ah yeah good idea, if only because it prevents duplicates.

shared-models/src/Publication/epub/Properties+EPUB.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+Properties.prototype.getContains = function() {
+  return this.otherProperties["contains"] || [];
+}
+
+Properties.prototype.getLayout = function() {
+  return this.otherProperties["layout"] || null
+}
```
mickael-menu, line 22: This is similar to the `Encryption` comment earlier, and concerns all of these helpers. I guess it's returning the raw JSON values here, so we need to parse them in a proper type (e.g. `EPUBLayout`) before returning them.

shared-models/src/Publication/presentation/Metadata+Presentation.ts
```diff
@@ -0,0 +1,19 @@
...
+declare module "../Metadata" {
+  export interface Metadata {
+    getPresentation: () => Presentation;
+  }
+}
+
+Metadata.prototype.getPresentation = function () {
+  return (this.otherMetadata && this.otherMetadata.json["presentation"])
+    ? new Presentation(this.otherMetadata.json["presentation"])
+    : new Presentation({});
```
mickael-menu: Related to this comment: https://github.com/readium/web/pull/3/files#r456409931. Here it looks good

shared-models/src/Publication/Link.ts
JayPanoz: Yes. That’s an [index signature](https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures).

shared-models/src/Publication/epub/Properties+EPUB.ts
mickael-menu: I just realized that it might work because the enums are defined as string unions. Please ignore my comment if it's the case 👍

shared-models/src/Publication/Metadata.ts
JayPanoz: Whoops sorry, that was an oversight.

shared-models/src/Format/Format.ts
mickael-menu: Okay, would it make sense to add a `format.equals(Format.epub())` function then?

Testing `mediaType.string` should be enough.

shared-models/src/Publication/Link.ts
mickael-menu: Yeah and it's more convenient when doing unit tests, because altering the set means that we don't have to check the order as well.

shared-models/src/Publication/Link.ts
mickael-menu: Interesting, thanks!

shared-models/src/Publication/Metadata.ts
JayPanoz: They’re working the other way around here, `parseSomething()` pops them itself since [deleting an object returns a boolean in JS.](https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures) 

That said, I’ve been wondering whether that should really be a class instead of a set of parse functions.

mickael-menu: Ha yes, sorry I missed the `pop` in `JSONDictionary`.

Personally I think it's fine (and OOer) as a wrapping class which operates on the dictionary, but then maybe at the end you can do `otherMetadata = jsonDictionary.json` to get the remaining properties as a simple `Map`, and discard the `JSONDictionary`

shared-models/src/Format/Format.ts
JayPanoz: That’s an option considering we only need shallow equality for `Format` and `mediaType.string` is a primitive so it will compare values out of the box, so at least that would provide sensible defaults for whom doesn’t want to deal with the issue.

shared-models/src/Publication/Encryption/Properties+Encryption.ts
JayPanoz: Properties should parse it into an object if the input (constructor) is a JSON string, but maybe I’m missing something given Encryption hasn’t been part of my uses cases for cloud so I thought it would be better addressed by someone who needed that.

shared-models/src/Publication/Metadata.ts
JayPanoz: That would be me trying to reconcile parsing + JSON string ([serialisation has a couple of rules to take into account](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)) as we were talking about the need of a parser client-side.

Then the difference is that `null` is the absence of value for a variable while `undefined` is the absence of variable itself in this context.

shared-models/src/Publication/Encryption/Properties+Encryption.ts
mickael-menu: For consistency, `otherProperties`, `otherMetadata` and `otherLocations` should always contain raw JSON value, even if the input source is not a JSON. Otherwise the caller needs to check what kind of value it gets for the helper and do the parsing manually. 

However you have a good point, encryption is not really a use case for web at the moment, so I think we can remove the `Encryption` type and leave reading apps implement it themselves through the model extensibility if really needed.

llemeurfr: there are good chances that the online protection we'll apply will use some sort of encryption. Therefore we can leave the property.

JayPanoz: Ah, I’ll have to double-check the issue is resolved before merging it then. Will do ASAP.

PR #8, opened 2021-07-29
Readium Shared Models recoded from scratch

bilal-uslu:


.github/workflows/CI-shared-model.yml
```diff
@@ -0,0 +1,42 @@
+name: CI-shared-models
```
mickael-menu: 👍 for adding a CI workflow

That raises the question: how should we handle the GitHub workflows for all the packages in the monorepo?

* A single workflow with one job per package (my preference)
* One workflow per package (like here)

There will be a lot of duplication, but maybe GitHub actions allow to reuse some of the steps among jobs.

.github/workflows/size-shared-model.yml
```diff
@@ -0,0 +1,14 @@
+name: size-shared-model
```
mickael-menu: What is this workflow used for?

.gitignore
```diff
@@ -0,0 +1,7 @@
+*.log
+.DS_Store
+node_modules
+dist
+example/build/jellybookslib.js
```
mickael-menu: We'll need to sort out every occurrence of Jellybooks specific mentions.

shared-models/package.json
```diff
@@ -1,19 +1,73 @@
 {
-  "name": "@readium/shared-web",
-  "version": "1.0.0",
+  "name": "@jellybooks/shared-test",
```
mickael-menu: Another question for a call: how to organize the packages in the repo?

`shared-models` doesn't match other platforms, and it won't necessarily contain only models so we could rename it to only `shared`.

However we could also embrace the monorepo and make many smaller packages, so I would split `shared-models` into:
* `opds` – all models and parsers related to OPDS
* `publication` or `pub`  – the shared `Publication` models as well as publication service definitions and their default implementation
* `mediatype` – the media type definitions and sniffers

This is the approach we are taking in `go-toolkit`.

shared-models/src/Publication/LocatorCollection.ts
```diff
@@ -0,0 +1,133 @@
...
+
+import { numberfromJSON } from '../util/JSONParse';
+import { Links } from './Link';
+import { LocalizedString } from './LocalizedString';
+import { Locator } from './Locator';
+
+/**
+ * Holds the metadata of a `LocatorCollection`.
+ */
+export class LocatorMetadata {
```
mickael-menu: If we can't have a nested `LocatorCollection.Metadata` type, then maybe we should rename this into `LocatorCollectionMetadata`? On first read, I was confused as I thought this was the metadata of a single `Locator` object.

shared-models/src/Publication/BelongsTo.ts
```diff
@@ -0,0 +1,51 @@
+/* Copyright 2021 Readium Foundation. All rights reserved.
```
mickael-menu: I know the project had some capitalized folders as well before this PR so it's not a comment for it in particular. But this doesn't seem to be a standard convention. Should we change everything to lowercase?

Screenshot 2021-08-05 at 10 03 44

shared-models/src/Publication/Link.ts
```diff
@@ -1,222 +1,330 @@
...
  */
 export class Link {
-
   /** URI or URI template of the linked resource. */
   public href: string;
 
   /** Indicates that a URI template is used in href. */
   public templated?: boolean;
-  
+
```
mickael-menu: We can put `Link` properties as `readonly` as well, unless you had a specific need in mind for mutability?

shared-models/src/Publication/Link.ts
```diff
@@ -1,222 +1,330 @@
...
   /** Title of the linked resource. */
   public title?: string;
-  
+
   /** Relation between the linked resource and its containing collection. */
-  public rels: Set<string>;
-  
+  public rels?: Set<string>;
```
mickael-menu: I guess having list/set properties as nullable is fine. On mobile we avoided nullability except for litteral values to simplify calling code:

```typescript
link.rels.contains("cover")
```

vs

```typescript
link.rels?.contains("cover") == true
```

shared-models/src/Publication/Link.ts
```diff
@@ -1,222 +1,330 @@
...
-    this.height = link.height;
-    this.width = link.width;
-    this.duration = link.duration;
-    this.bitrate = link.bitrate;
-    this.languages = link.language;
-    this.alternates = link.alternate ? new Links(link.alternate) : new Links([]);
-    this.children = link.children ? new Links(link.children) : new Links([]);
-    this.mediaType = link.type ? new MediaType(link.type) : undefined;
-    this.templateParameters = this.getTemplateParameters();
+  public getMediaType(): MediaType {
```
mickael-menu: The sniffing of a  `Link` media type will be really fast since it's only comparing `type`, so I think a getter would be nicer here to hide the fact it's a computed property:

```suggestion
  public get mediaType(): MediaType {
```

I like to use this convention: use a property/getter when:

* it's semantically a field
* it doesn't have any side effect
* it returns relatively quickly and so can be called from the UI thread

Otherwise I use a method.

shared-models/src/Publication/Link.ts
```diff
@@ -1,222 +1,330 @@
...
+      ? MediaType.parse({ mediaType: this.type })
+      : MediaType.BINARY;
   }
 
   /** Computes an absolute URL to the link, relative to the given `baseURL`.
    *  If the link's `href` is already absolute, the `baseURL` is ignored.
    */
-  public toAbsoluteHREF(baseUrl: string): string {
-    return new URL(this.href, baseUrl).href;
+  public toAbsoluteHREF(baseUrl?: string): string | undefined {
```
mickael-menu: We name it `toURL()` on other platforms

```suggestion
  public toURL(baseUrl?: string): string | undefined {
```

shared-models/src/Publication/Link.ts
```diff
@@ -1,222 +1,330 @@
...
   }
 
   /** List of URI template parameter keys, if the `Link` is templated. */
-  private getTemplateParameters(): Set<string> {
-    if (!this.templated) {
-      return new Set();
-    } else {
-      return new URITemplate(this.href).parameters;
-    }
+  public getTemplateParameters(): Set<string> {
```
mickael-menu:
```suggestion
  public get templateParameters(): Set<string> {
```

shared-models/src/Publication/Link.ts
```diff
@@ -1,222 +1,330 @@
...
+    link.properties = link.properties
+      ? link.properties?.add(properties)
+      : new Properties(properties);
+    return link;
+  }
+
+  /**
+   * Creates a [Locator] from a reading order [Link].
+   */
+  public toLocator(): Locator {
```
mickael-menu:
```suggestion
  public get locator(): Locator {
```

shared-models/src/Publication/Locator.ts
```diff
@@ -1,65 +1,246 @@
...
-  partialCfi?: string;
-  domRange?: IDOMRange;
+  // cssSelector?: string;
+  // partialCfi?: string;
+  // domRange?: IDOMRange;
 }
 
-/** https://github.com/readium/architecture/tree/master/locators */
-export interface ILocator {
+export class Text {
```
mickael-menu:
```suggestion
export class LocatorText {
```

shared-models/src/Publication/Locator.ts
```diff
@@ -1,65 +1,246 @@
...
-}
-
-/** One or more alternative expressions of the location.
- *  https://github.com/readium/architecture/tree/master/models/locators#the-location-object
+/**
+ * One or more alternative expressions of the location.
+ * https://github.com/readium/architecture/tree/master/models/locators#the-location-object
  */
-export interface ILocations {
+export class Locations {
```
mickael-menu: If we can't have a `Locator.` namespace in Typescript, lets prefix the types.

```suggestion
export class LocatorLocations {
```

shared-models/src/Publication/Locator.ts
```diff
@@ -1,65 +1,246 @@
...
   /** Contains one or more fragment in the resource referenced by the `Locator`. */
-  fragments: Array<string>;
-  
+  public fragments: Array<string>;
```
mickael-menu: The `Locator` properties can be `readonly` as well.

shared-models/src/Publication/Locator.ts
```diff
@@ -1,65 +1,246 @@
...
+    }
+    return json;
+  }
+
+  //TODO :Extend
   /** otherLocators currently in use in Thorium/R2D2BC */
-  cssSelector?: string;
-  partialCfi?: string;
-  domRange?: IDOMRange;
+  // cssSelector?: string;
```
mickael-menu: Yeah I think we can remove this altogether. It can always be implemented in `otherLocations`.

shared-models/src/Publication/Locator.ts
```diff
@@ -1,65 +1,246 @@
...
+    if (this.title !== undefined) json.title = this.title;
+    if (this.locations) json.locations = this.locations.serialize();
+    if (this.text) json.text = this.text.serialize();
+    return json;
+  }
+
+  /**
+   * Shortcut to get a copy of the [Locator] with different [Locations] sub-properties.
+   */
+  public copyWithLocations(values: any): Locator {
```
mickael-menu: I don't know much TypeScript, but feel free to adapt this kind of code to what you would usually find in TypeScript native libs. If there's an easier way to modify an immutable object.

shared-models/src/Publication/Manifest.ts
```diff
@@ -57,8 +123,110 @@ export class Manifest {
...
+      }
+      return href;
+    }
+    return;
+  }
+
+  /**
+   * Sets the URL where this [Publication]'s RWPM manifest is served.
+   */
+  public setSelfLink(href: string): void {
```
mickael-menu: I think this won't be needed for Readium Web, it was because of the internal HTTP server on mobile.

shared-models/src/Publication/Manifest.ts
```diff
@@ -57,8 +123,110 @@ export class Manifest {
...
+
+    const shortHref = parts[0];
+
+    return this.linkWithHref(shortHref);
+  }
+
+  /** The URL where this publication is served, computed from the `Link` with `self` relation.
+   *  e.g. https://provider.com/pub1293/manifest.json gives https://provider.com/pub1293/
+   */
+  public getBaseURL(): string | undefined {
```
mickael-menu:
```suggestion
  public get baseURL(): string | undefined {
```

shared-models/src/Publication/Metadata.ts
```diff
@@ -1,139 +1,329 @@
...
+    return json;
+  }
+
+  /**
+   * Computes a [ReadingProgression] when the value of [readingProgression] is set to
+   * auto, using the publication language.
+   *
+   * See this issue for more details: https://github.com/readium/architecture/issues/113
    */
   public effectiveReadingProgression(): ReadingProgression {
```
mickael-menu:
```suggestion
  public get effectiveReadingProgression(): ReadingProgression {
```

shared-models/src/Publication/BelongsTo.ts
```diff
@@ -0,0 +1,51 @@
...
+ * available in the LICENSE file present in the Github repository of the project.
+ */
+
+import { Contributors } from './Contributor';
+
+/**
+ * BelongsTo Object for the Readium Web Publication Manifest.
+ * https://readium.org/webpub-manifest/schema/contributor-object.schema.json
+ */
+export class BelongsTo {
```
mickael-menu: I'm not sure we need a dedicated `BelongsTo` object, but maybe it's easier in the TypeScript codebase.

On mobile we have `Metadata.belongsTo: Map<String, [Collection]>`.

shared-models/src/Publication/BelongsTo.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Map of [BelongsTo] items.
+   */
+  public readonly items: Map<string, Contributors>;
```
mickael-menu: The term `Contributor` is a bit confusing in this context. Since it's basically the same object, we added a typealias on mobile. Is is something feasible with TypeScript?

```
public typealias Collection = Contributor
```

shared-models/src/Publication/Publication.ts
```diff
@@ -1,103 +1,101 @@
...
+    return list && list.length > 0 ? list[0].links : undefined;
   }
 
-  /** Finds the first link with the given relation in the publication's links. */
-  public linkWithRel(rel: string): Link | null {
-    return this.manifest.linkWithRel(rel);
+  /**
+   * Sets the URL where this [Publication]'s RWPM manifest is served.
+   */
+  public setSelfLink(href: string): void {
```
mickael-menu: I think this won't be needed for Readium Web.

shared-models/src/Publication/Publication.ts
```diff
@@ -1,103 +1,101 @@
...
+   * Finds the first [Link] having the given [rel] in the publications's links.
+   */
+  public linkWithRel(rel: string): Link | undefined {
+    return this.manifest.linkWithRel(rel);
+  }
+
+  /**
+   * Returns Type of publication
+   */
+  public getType(): PublicationType {
```
mickael-menu: `PublicationType` was some legacy code in mobile, and we will deprecate it soon. So I think we don't need it for Web.

shared-models/src/Publication/Publication.ts
```diff
@@ -1,103 +1,101 @@
...
    */
-  public baseURL(): string | null {
-    const selfLink = this.manifest.links.find(el => el.rels.has("self"));
-    if (selfLink) {
-      return selfLink.href;
-    } else {
-      return null;
-    }
-  };
+  public getBaseURL(): string | undefined {
```
mickael-menu:
```suggestion
  public get baseURL(): string | undefined {
```

shared-models/src/Publication/PublicationCollection.ts
```diff
@@ -0,0 +1,144 @@
...
+      links,
+      subcollections: subcollections?.size ? subcollections : undefined,
+    });
+  }
+
+  /**
+   * Parses a map of [PublicationCollection] indexed by their roles from its RWPM JSON representation.
+   *
+   */
+  public static collectionsFromJSON(
```
mickael-menu: Maybe `deserializeCollections()`?

shared-models/src/Publication/encryption/Properties.ts
```diff
@@ -0,0 +1,18 @@
...
+     * Indicates that a resource is encrypted/obfuscated and provides relevant information for
+     * decryption.
+     */
+    getEncryption(): Encryption | undefined;
+  }
+}
+
+Properties.prototype.getEncryption = function(): Encryption | undefined {
+  return Encryption.deserialize(this.otherProperties.encrypted);
+};
```
mickael-menu: Extensions seem really easy with TypeScript as well!

shared-models/src/Publication/epub/EpubLayout.ts
```diff
@@ -1,10 +1,10 @@
...
  * available in the LICENSE file present in the Github repository of the project.
  */
 
 /** Hint about the nature of the layout for the linked resources. */
-export enum EPUBLayout {
-  fixed = "fixed",
-  reflowable = "reflowable"
-}
\ No newline at end of file
+export enum EpubLayout {
```
mickael-menu: Isn't the convention to keep accronyms as uppercase?

```suggestion
export enum EPUBLayout {
```

shared-models/src/Publication/epub/Properties.ts
```diff
@@ -0,0 +1,28 @@
...
+// EPUB extensions for link [Properties].
+// https://readium.org/webpub-manifest/schema/extensions/epub/properties.schema.json
+
+declare module '../Properties' {
+  export interface Properties {
+    /**
+     * Identifies content contained in the linked resource, that cannot be strictly identified using a
+     * media type.
+     */
+    getContains(): Set<string> | undefined;
```
mickael-menu: Is it possible to have getters as an extension?

shared-models/src/util/mediatype/MediaType.ts
```diff
@@ -0,0 +1,542 @@
...
+      this.matchesAny(
+        MediaType.ZIP,
+        MediaType.LCP_PROTECTED_AUDIOBOOK,
+        MediaType.LCP_PROTECTED_PDF
+      ) || this.structuredSyntaxSuffix === '+zip'
+    );
+  }
+
+  /** Returns whether this media type is structured as a JSON file. */
+  public isJson(): boolean {
```
mickael-menu:
```suggestion
  public isJSON(): boolean {
```

shared-models/src/util/mediatype/MediaType.ts
```diff
@@ -0,0 +1,542 @@
...
+    return (
+      this.matchesAny(MediaType.JSON) || this.structuredSyntaxSuffix === '+json'
+    );
+  }
+
+  /** Returns whether this media type is of an OPDS feed. */
+  public isOpds(): boolean {
```
mickael-menu:
```suggestion
  public isOPDS(): boolean {
```

shared-models/src/util/mediatype/MediaType.ts
```diff
@@ -0,0 +1,542 @@
...
+        MediaType.OPDS1_ENTRY,
+        MediaType.OPDS2,
+        MediaType.OPDS2_PUBLICATION,
+        MediaType.OPDS_AUTHENTICATION
+      ) || this.structuredSyntaxSuffix === '+json'
+    );
+  }
+
+  /** Returns whether this media type is of an HTML document. */
+  public isHtml(): boolean {
```
mickael-menu:
```suggestion
  public isHTML(): boolean {
```

shared-models/src/util/mediatype/MediaType.ts
```diff
@@ -0,0 +1,542 @@
...
+    return false;
+  }
+
+  /** Checks the MediaType equals another one (comparing their string) */
+  public equals(other: MediaType): boolean {
+    return this.string === other.string;
+  }
+
+  /** Returns whether this media type is structured as a ZIP archive. */
+  public isZIP(): boolean {
```
mickael-menu:
```suggestion
  public get isZIP(): boolean {
```

mickael-menu (changes requested):
Overall great changes!

.github/workflows/CI-shared-model.yml
bilal-uslu: There is a filter,
for the specific package,
but still i would prefer keep shared models and navigator package in seperate repos, It does not make sense to me have all in a single repo, since there is no significance advantage

name: CI-shared-models
on: 
  push:
    paths:
      - 'shared-models/**'
      - '.github/workflows/CI-shared-models.yml'

.github/workflows/size-shared-model.yml
bilal-uslu: to keep the created package in a specific size, too prevent publishing of a huge size of a library

shared-models/package.json
bilal-uslu: if we are going to publish it on npm as a package then we should keep them together. Otherwise there may be too much npm packages

.github/workflows/CI-shared-model.yml
mickael-menu: We've already decided to go with monorepos for all the reasons explained in this discussion: https://github.com/readium/architecture/discussions/136

Believe me we have the experience of multiple repos with the Readium architecture and it's a pain to maintain.

mickael-menu: Although feel free to chime in in this discussion after taking a look at the arguments with some counter-arguments specific to Readium Web.

mickael-menu: By the way, I think clarifying what will the navigator provide could help too. We want to provide a very simple API that has no user interface. Basically something like `Navigator(containerElement)`, and then it will create the iframes and handle the scrolling inside the container. There won't be any progression bar or "bookmark page" button, for example. And the navigator won't depend on any high-level UI framework (React, Vue.js, etc.)

Then, each implementer will have their own repo to create the actual reader user interface using whatever UI framework they want and using Readium's navigator only for the "rendering" component/view.

Readium will also provide a set of interfaces that Navigator implementations will need to implement. This way a reading app using these interfaces can switch to a different navigator implementation (maybe not provided by Readium) without changing any of their app code.

Does having the navigator interfaces and default implementations in the mono-repo makes more sense for you in this context?

bilal-uslu: Ok. It is more clear now.

.gitignore
bilal-uslu: removed

shared-models/package.json
bilal-uslu: changed shared-models to shared

shared-models/src/Publication/LocatorCollection.ts
bilal-uslu: renamed

shared-models/src/Publication/BelongsTo.ts
bilal-uslu: changed

shared-models/src/Publication/Link.ts
bilal-uslu: done

shared-models/src/Publication/Link.ts
bilal-uslu: for serialization and deserialization it is better to keep like that. by the way it is a reasonable usage for typescript and javascript

shared-models/src/Publication/Link.ts
bilal-uslu: done

shared-models/src/Publication/Link.ts
bilal-uslu: done

shared-models/src/Publication/Link.ts
bilal-uslu: done

shared-models/src/Publication/Link.ts
bilal-uslu: done

shared-models/src/Publication/Locator.ts
bilal-uslu: renamed

shared-models/src/Publication/Locator.ts
bilal-uslu: renamed

shared-models/src/Publication/Locator.ts
bilal-uslu: done

shared-models/src/Publication/Locator.ts
bilal-uslu: removed

shared-models/src/Publication/Locator.ts
bilal-uslu: simplified

shared-models/src/Publication/Manifest.ts
bilal-uslu: removed

shared-models/src/Publication/Manifest.ts
bilal-uslu: done

shared-models/src/Publication/Metadata.ts
bilal-uslu: done

shared-models/src/Publication/BelongsTo.ts
bilal-uslu: yes. this way is easier

shared-models/src/Publication/BelongsTo.ts
bilal-uslu: yes, we can. I have aliased it to Collection

shared-models/src/Publication/Publication.ts
bilal-uslu: removed

shared-models/src/Publication/Publication.ts
bilal-uslu: removed

shared-models/src/Publication/Publication.ts
bilal-uslu: done

shared-models/src/Publication/PublicationCollection.ts
bilal-uslu: good point, renamed,

shared-models/src/Publication/encryption/Properties.ts
bilal-uslu: yes exactly, to setup this part was a bit painful, but looks good now

shared-models/src/Publication/epub/EpubLayout.ts
bilal-uslu: renamed

shared-models/src/Publication/epub/Properties.ts
bilal-uslu: yes, done

shared-models/src/util/mediatype/MediaType.ts
bilal-uslu: renamed

shared-models/src/util/mediatype/MediaType.ts
bilal-uslu: renamed

shared-models/src/util/mediatype/MediaType.ts
bilal-uslu: renamed

shared-models/src/util/mediatype/MediaType.ts
bilal-uslu: done

shared-models/src/Publication/Link.ts
mickael-menu: Alright

shared-models/src/publication/Publication.ts
```diff
@@ -0,0 +1,81 @@
...
+import { PublicationCollection } from './PublicationCollection';
+
+export enum PublicationType {
+  EPUB = 'EPUB',
+  CBZ = 'CBZ',
+  FXL = 'FXL',
+  WEBPUB = 'WEBPUB',
+  AUDIO = 'AUDIO',
+  DiViNa = 'DiViNa',
+}
```
mickael-menu: We can remove this enum now.

shared-models/src/publication/Locator.ts
```diff
@@ -0,0 +1,225 @@
...
+      this.otherLocations.forEach((value, key) => (json[key] = value));
+    }
+    return json;
+  }
+}
+
+export class LocatorText {
+  public after?: string;
+  public before?: string;
+  public highlight?: string;
```
mickael-menu:
```suggestion
  public readonly after?: string;
  public readonly before?: string;
  public readonly highlight?: string;
```

shared-models/src/publication/Locator.ts
```diff
@@ -0,0 +1,225 @@
...
+  public type: string;
+
+  /** The title of the chapter or section which is more relevant in the context of this locator. */
+  public title?: string;
+
+  /** One or more alternative expressions of the location. */
+  public locations: LocatorLocations;
+
+  /** Textual context of the locator. */
+  public text?: LocatorText;
```
mickael-menu:
```suggestion
  public readonly href: string;

  /** The media type of the resource that the Locator Object points to. */
  public readonly type: string;

  /** The title of the chapter or section which is more relevant in the context of this locator. */
  public readonly title?: string;

  /** One or more alternative expressions of the location. */
  public readonly locations: LocatorLocations;

  /** Textual context of the locator. */
  public readonly text?: LocatorText;
```

shared-models/src/util/mediatype/MediaType.ts
```diff
@@ -0,0 +1,542 @@
...
+    return this.type === 'audio';
+  }
+
+  /** Returns whether this media type is of a video clip. */
+  public get isVideo(): boolean {
+    return this.type === 'video';
+  }
+
+  /** Returns whether this media type is of a Readium Web Publication Manifest. */
+  public get isRwpm(): boolean {
```
mickael-menu:
```suggestion
  public get isRWPM(): boolean {
```

mickael-menu (changes requested):
Mostly fine, there's a couple of suggestions left.

Also could you rename the `shared-models` folder into just `shared`? Thanks.

shared-models/src/publication/Publication.ts
bilal-uslu: removed

shared-models/src/publication/Locator.ts
bilal-uslu: done

shared-models/src/publication/Locator.ts
bilal-uslu: done

shared-models/src/util/mediatype/MediaType.ts
bilal-uslu: renamed

mickael-menu (approved):

mickael-menu:
I'm not sure the GitHub actions will trigger until we merge this PR in, as you are not maintainer on the repo.

shared/src/publication/Link.ts
```diff
@@ -0,0 +1,330 @@
...
+  public filterByMediaType(mediaType: string): Array<Link> {
+    const predicate = (el: Link) => el.mediaType.matches(mediaType);
+    return this.items.filter(predicate);
+  }
+
+  /** Finds all the links matching any of the given media types. */
+  public filterByMediaTypes(mediaTypes: Array<string>): Array<Link> {
+    const predicate = (el: Link) => {
+      for (const mediaType of mediaTypes) {
+        return el.mediaType.matches(mediaType);
```
dysbulic: I'm pretty sure this will only check the first element of the array. You need to only `return` if it's found.

```suggestion
        if (el.mediaType.matches(mediaType)) {
          return true;
        }
```

A quicker solution would be to replace the loop with:

```suggestion
        return mediaTypes.includes(el.mediaType);
```

shared/src/publication/Locator.ts
```diff
@@ -0,0 +1,225 @@
...
+        progression !== undefined && progression >= 0 && progression <= 1
+          ? progression
+          : undefined,
+      totalProgression:
+        totalProgression !== undefined &&
+        totalProgression >= 0 &&
+        totalProgression <= 1
+          ? totalProgression
+          : undefined,
+      position: position !== undefined && position > 0 ? position : undefined,
```
dysbulic, line 81: `undefined > 0` is `false`, so you can just put:

```suggestion
      position: position > 0 ? position : undefined,
```

shared/src/publication/Link.ts
bilal-uslu: good point. fixed. thanks

shared/src/publication/Locator.ts
```diff
@@ -0,0 +1,225 @@
...
+      text: LocatorText.deserialize(json.text),
+    });
+  }
+
+  /**
+   * Serializes a [Link] to its RWPM JSON representation.
+   */
+  public serialize(): any {
+    const json: any = { href: this.href, type: this.type };
+    if (this.title !== undefined) json.title = this.title;
```
dysbulic, line 207: As a pattern in your serializer functions, it is safe to assign `undefined` to variables that are already `undefined`, so you can remove many of your `if` guards.

```suggestion
    json.title = this.title;
```

shared/src/publication/Manifest.ts
```diff
@@ -0,0 +1,232 @@
...
+  public get baseURL(): string | undefined {
+    const selfLink = this.links.items.find(
+      el => el.rels && el.rels.has('self')
+    );
+    if (selfLink) {
+      let href = selfLink.href;
+      if (href) {
+        const li = href.lastIndexOf('/');
+        const lastpart =
+          li === -1 ? undefined : href.substring(href.lastIndexOf('/') + 1);
```
dysbulic:
```suggestion
          li === -1 ? undefined : href.substring(li + 1);
```

shared/src/publication/Manifest.ts
```diff
@@ -0,0 +1,232 @@
...
+
+    return this.linkWithHref(shortHref);
+  }
+
+  /** The URL where this publication is served, computed from the `Link` with `self` relation.
+   *  e.g. https://provider.com/pub1293/manifest.json gives https://provider.com/pub1293/
+   */
+  public get baseURL(): string | undefined {
+    const selfLink = this.links.items.find(
+      el => el.rels && el.rels.has('self')
```
dysbulic, line 197:
```suggestion
      el => el.rels?.has?.('self')
```

shared/src/publication/Manifest.ts
```diff
@@ -0,0 +1,232 @@
...
+    );
+    if (selfLink) {
+      let href = selfLink.href;
```
dysbulic, line 200: Remove the `if (selfLink) {` statement and just have:

```suggestion
      let href = selfLink?.href;
```

shared/src/publication/Metadata.ts
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /**
+   * Serializes a [Metadata] to its RWPM JSON representation.
+   */
+  public serialize(): any {
+    const json: any = { title: this.title.serialize() };
+    if (this.typeUri !== undefined) json['@type'] = this.typeUri;
+    if (this.identifier !== undefined) json.identifier = this.identifier;
+    if (this.subtitle) json.subtitle = this.subtitle.serialize();
```
dysbulic, line 253: Another pattern that allows you to remove more of your `if` guards is the use of the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator:

```suggestion
    json.subtitle = this.subtitle?.serialize();
```

shared/src/publication/Publication.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.manifest.linkWithHref(href);
+  }
+
+  /**
+   * Returns the [links] of the first child [PublicationCollection] with the given role, or an
+   * empty list.
+   */
+  public linksWithRole(role: string): Links | undefined {
+    const list = this.subcollections?.get(role);
+    return list && list.length > 0 ? list[0].links : undefined;
```
dysbulic, line 58:
```suggestion
    return list?.[0]?.links;
```

shared/src/opds/Acquisition.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(values: { type: string; children?: Array<Acquisition> }) {
+    this.type = values.type;
+    this.children = values.children;
+  }
+
+  /**
+   * Parses a [Acquisition] from its RWPM JSON representation.
+   */
+  public static deserialize(json: any): Acquisition | undefined {
+    if (!(json && json.type)) return;
```
dysbulic, line 29:
```suggestion
    if (!json?.type) return;
```

shared/src/util/JSONParse.ts
```diff
@@ -0,0 +1,31 @@
+/** Parses the given array (or undefined it it's not an array) */
+export function arrayfromJSON(json: any): Array<any> | undefined {
+  return json && json instanceof Array ? json : undefined;
```
dysbulic, line 3: `undefined instanceof Array` (or `null` or `0` or any falsey value) is `false`, so you just need: 

```suggestion
  return json instanceof Array ? json : undefined;
```

shared/src/util/URITemplate.ts
```diff
@@ -39,28 +47,37 @@ export class URITemplate {
    */
   public expand(parameters: { [param: string]: string }): string {
     const expandSimpleString = (string: string): string => {
-      return string.split(",")
+      return string
+        .split(',')
         .map(parameter => {
-          return parameters[parameter] || "";
+          const parameterValue = parameters[parameter];
```
dysbulic, line 53:
```suggestion
          const parameterValue = parameters[parameter] ?? '';
```

shared/src/util/URITemplate.ts
```diff
@@ -39,28 +47,37 @@ export class URITemplate {
...
+          return parameterValue ? encodeURIComponent(parameterValue) : '';
```
dysbulic, line 54:
```suggestion
          return encodeURIComponent(parameterValue);
```

PR #9, opened 2021-08-27
Readium Web Navigator Implementation

bilal-uslu:
Readium Web Navigator Implementation


bilal-uslu:
please note that, size-navigator / size (pull_request) is failed, which is normal, because we are going to add a pipeline at the first time

chocolatkey:
What caused you to need to increase the size limit from 10KB to 50KB?

bilal-uslu:
> What caused you to need to increase the size limit from 10KB to 50KB?

10 KB is initial value and it is 50 KB for shared already. The bundle size of this library is 14 KB now. If it goes over 50KB, it will give you a warning. Any size larger than 15 KB works at the moment. Here I choose 50 KB, which seems reasonable to me. This is to avoid unexpected packet sizes.

navigator/src/ui/Color.ts
```diff
@@ -0,0 +1,30 @@
...
+  public toString() {
+    return `rgba(${[this.r, this.g, this.b, this.a].join(',')})`;
+  }
+
+  public static get black(): Color {
+    return new Color(0, 0, 0, 1);
+  }
+
+  public static get white(): Color {
+    return new Color(0, 0, 0, 0);
```
chocolatkey: I think white should not be transparency, but RGBA all at max

bilal-uslu: Right. Fixed.

PR #10, opened 2022-03-23
Restructure package layout, add minimum required for position list fetching

chocolatkey:


shared/src/fetcher/HttpFetcher.ts
```diff
@@ -0,0 +1,114 @@
+import { Link } from '../publication/Link';
+import { Fetcher } from './Fetcher';
+import { NumberRange, Resource } from './Resource';
+
+// Fetches remote resources through HTTP.
+export class HttpFetcher implements Fetcher {
+  private readonly baseUrl?: string;
+  // TODO HttpClient? Is it even necesary when we have fetch()?
```
mickael-menu: Yes probably not in TypeScript.

shared/src/fetcher/Resource.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+export interface Resource {
+  link(): Promise<Link>;
+  length(): Promise<number | undefined>; // TODO make try?
+  read(range?: NumberRange): Promise<Uint8Array | undefined>;
+  readAsString(): Promise<string | undefined>;
+  readAsJSON(): Promise<unknown | undefined>;
+  readAsXML(): Promise<XMLDocument | undefined>;
```
mickael-menu: It looks like TypeScript doesn't support default body for interface methods?

If that's the case, and there's no way to add extensions to any `Resource` object, then we can remove these helpers to avoid having to reimplement them for all `Resource` implementations.

Here are some alternatives, what do you think?

* Using an abstract class instead of an `interface` which implement these helpers. Usually resources are their own object and don't implement other interfaces.
* Having free-floating functions like `readAsString(resource)`.

shared/src/publication/Publication.ts
```diff
@@ -69,4 +76,25 @@ export class Publication {
...
+      'application/vnd.readium.position-list+json'
+    );
+    if (positionListLink === undefined) return [];
+    const positionListJSON = (await this.get(
+      positionListLink
+    ).readAsJSON()) as { positions: unknown[] };
+    return (positionListJSON['positions'] as unknown[]) // Get the array for the positions key
+      .map(pos => Locator.deserialize(pos)) // Parse locators
+      .filter(l => l !== undefined) as Locator[]; // Filter out failures
+  }
```
mickael-menu, line 91: We'll need to delegate to a `PositionsService` interface eventually to be able to swap implementations. For example if a server doesn't have a position list, a web reader might generate its own.

However it's not blocking for this PR.

mickael-menu (changes requested):

shared/src/fetcher/HttpFetcher.ts
chocolatkey: I changed my mind, included an option to specify a client. Because in some situations there are people who want e.g. an authorized `fetch`, or a `fetch` that retries things like https://www.npmjs.com/package/fetch-retry

shared/src/fetcher/Resource.ts
chocolatkey: I went with the free-floating functions for the go toolkit, as you can see here https://github.com/readium/go-toolkit/blob/master/pkg/fetcher/fetcher_file.go#L209-L221 , it works fine. Do note that in the case of HttpResource, they're intentionally overridden because the native fetch has convenient ways of decoding the response.
So if I were to implement e.g. FileFetcher then I'd add the free-floating functions, but for Http alone they're unnecessary. @mickael-menu let me know if this is OK

shared/src/publication/Publication.ts
chocolatkey: Yes, that's the idea. Not implemented yet for simplicity's sake

shared/src/fetcher/Resource.ts
mickael-menu: That's fine by me, then can you just remove these functions from the `Resource` interface? Thanks.

chocolatkey: @mickael-menu So you're sure you want this to be different vs. the go/kotlin/swift toolkits?

mickael-menu: The most idiomatic solution for the language is the best, so that's fine if it's not a direct translation from the other toolkits. Do you know a good alternative that doesn't require all implementations of `Resource` to reimplement these functions?

chocolatkey: Personally I'm prioritizing keeping the toolkits similar, because it will (hopefully) make it easier to keep them in sync in the long run. I was suggesting that every `Resource` implementation re-implement it with the stub, like here https://github.com/readium/go-toolkit/blob/master/pkg/fetcher/fetcher_file.go#L210 .
Ultimately though, you are more experienced with the toolkits and their architecture, so if you think it's fine to have this sort of difference it's fine with me too.

mickael-menu: I agree for the core types and APIs, but the helpers are born from specific platform needs and don't necessarily need to be the same everywhere. So there's not a strong requirement to keep the helpers in sync until the need arise while using the toolkit itself. However, `readAsString/JSON/XML` are all very useful for the web toolkit.

The problem with requiring these helpers in the `Resource` interface is that:

* We will need to update all the implementations if we add a new helper, and this will require a new major semantic version release as it will break custom app implementations.
* This is not [extensible by implementers](https://elizarov.medium.com/extension-oriented-design-13f4f27deaee) wanting to add their own helpers. I'd rather have a way to extend the toolkit that is consistent with internal and third-party extensions.

I think using an abstract class instead of an interface would solve all these problems, as an implementer should be able to use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) to add more helpers.

Alternatively, if you want to keep the interface, we could add a proxy `HelperResource` class which has all the helpers and can be extended with module augmentation. Then you would use:

```typescript
HelperResource(resource).readAsXML()
```

Or

```typescript
ResourceReader(resource).asXML()
```

chocolatkey: @mickael-menu how about now? https://github.com/readium/ts-toolkit/pull/10/commits/44d94739ea750f4f86689a8bb6ac76d55eb4feb4

mickael-menu: Looking good, thank you! Could you take a look at the failed check? Then we can merge if you're fine with it.

chocolatkey: I'll address that in another PR when I move to yarn workspaces, it's just due to re-arranging of code

PR #12, opened 2022-05-04
Navigator + navigator-html-injectables implementation

chocolatkey:
# Changelog

- Switched to using [pnpm](https://pnpm.io/) for package management, with workspaces.
- Added (Adding) the [`navigator-html-injectables`](https://github.com/readium/ts-toolkit/tree/navigator-html/navigator-html-injectables/src) package which can be used either inside a reflowable (X)HTML or outside (in a javascript environment) to provide access and control over a resource from a navigator on any modern browser or embedded browser frame. This is a replacement for the javascript stubs found in the mobile SDKs, such as [this](https://github.com/readium/kotlin-toolkit/tree/develop/readium/navigator/src/main/assets/_scripts/src).
    - Implements various enhancements to resources, such as handling of column-based horizontal reading. This is done using the concept of dynamically loadable "Modules", which can be extended, enabled or disabled on the fly.
    The modules can use the communication pipeline to send and receive messages (commands, logs etc.) from the navigator.
    - Provides a communication pipeline using built-in browser primitives for sending cross-frame messages. A receiver can be [implemented](https://github.com/readium/ts-toolkit/blob/navigator-html/navigator/src/epub/frame/FrameComms.ts) in a parent frame on the web, or natively on mobile platforms.
    Once the communication channel is set up, it is used to send events, and receive commands, which are defined [here](https://github.com/readium/ts-toolkit/blob/navigator-html/navigator-html/src/comms/keys.ts). It also acts as a way to receive logs from the frame, for debugging purposes.
- Added (Adding) the [`navigator`](https://github.com/readium/ts-toolkit/tree/navigator-html/navigator/src) package, providing a navigator for web platforms based on the readium Navigator spec. Work is primarily based on the [kotlin toolkit](https://github.com/readium/kotlin-toolkit/tree/develop/readium/navigator/src/main/java/org/readium/r2/navigator), but adapter for the needs of the web. At the moment, this only includes the Reflowable EPUB navigator.
The reflowable EPUB navigator lets the implementer specify a containing HTML element, in which the navigator "mounts" itself similar to a JS framework. It then creates and destroys iframe elements within to display content. It is important that any outside JS framework not meddle with the contents of this element. You can, however, apply CSS styles using `.readium-navigator-iframe`, like this:
```css
.readium-navigator-iframe {
  width: 100%;
  height: 100%;
}
```

Note: the navigator is not complete in terms of the public API functions that are expected of a readium navigator, this will be updated as the functionality is added.

# Review/Implementor Notes
(that should probably be a part of documentation in the future)

## General

Expect builds to fail or be incomplete. Along with the switch to Yarn PnP, the move to [Yarn Workspaces](https://yarnpkg.com/features/workspaces/) (read this for more info) will be made, allowing for the packages to have separate build processes and dependencies, but depend directly on each other.
Originally, the idea for this repo seems to have been to have one single build for everything, but with the advent of packages such as `navigator-html`, it's become evident that we need to split up builds. Trying to build completely separate bundles with different dependencies and browser support requirements is a pain, it's much easier to just have them be separate packages. And tsdx, which was originally used to simplify the building of the repo, has unfortunately slowly fallen behind trends (as well as being inadequate for certain build scenarios now needed), and so a build system inspired by what tsdx originally provided (ts, jest etc.) will be implemented.

## navigator-html package

- You should interact with this entire package through the `Loader` class. As with many things related to this package, a working example of interacting with this package in the `navigator` package, such as [here](https://github.com/readium/ts-toolkit/blob/navigator-html/navigator/src/epub/frame/FrameManager.ts).

### Communication ("Comms")

Message  data sent over the communications channel are simple JSON objects with this structure:
```ts
export interface CommsMessage {
    _readium: number; // Sanity/version-checking field
    _channel: string; // Channel ID
    id?: string; // Optional (but recommended!) unique identifier
    strict?: boolean; // Whether or not the event *must* be handled by the receiver
    key: CommsEventKey | CommsCommandKey; // The "key" for identification to the listener
    data: unknown; // The data to be sent to the module
}
```
An example message could look like this:
```json
{
    "_readium": 1,
    "_channel": "11f714d8-4fee-47de-94ab-1bc971a61529",
    "id": "1927-x7jbpd-13c5hmd",
    "data": [
        "--USER__appearance",
        "readium-night-on"
    ],
    "key": "set_property",
    "strict": false
}
```

The high-level overview for establishing communication with the IFrame ("IFrame" can be substituted with "Webview"):

1. `Comms` class is constructed in the IFrame. It starts listening for messages from anywhere.
2. Navigator sends a `_ping` message to the IFrame, using its `origin` (on the web, blob:xxx) to ensure it is the only recipient, with data that looks like this:
```json
{
    "_readium": 1,
    "_channel": "11f714d8-4fee-47de-94ab-1bc971a61529",
    "id": "1917-b4ciqy-10fdb1g",
    "key": "_ping",
    "strict": false
}
```
3. The IFrame receives this _ping message and restricts any further messages to those coming from the same origin and using the same channel ID (from the `_channel` key). The source of the message is also stored, and used to then send any future messages along with the relevant channel ID and origin. Overall, this helps enhance security and prevent accidental message propagation to and from the wrong IFrame.
4. The IFrame unlocks normal communication, and sends a `_pong` message back to the Navigator.
5. The Navigator receives the `_pong` and is now ready for normal communication as well.

Note that since module initialization can happen before/at the same time as this "handshake", the IFrame stores a buffer of unsent `log` messages originating from the modules, which are then all sent after step 4. This is the only type of message that modules should attempt to send of their own initiative during initialization.

Note that both the channel ID and the message ID are random strings. The `_channel` ID *should* be a secure random (on the web, using `window.crypto.randomUUID()`).
The message `id` doesn't have to be secure, but it is recommended that it be sufficiently random as to avoid collisions. A more important reason not to use a secure random, and why the IFrame does not do so, is because the performance of messages is critical to ensuring a smooth reading experience, and secure randoms tend to take longer to generate, which would slow down communication. The IDs used by the IFrame and web implementation are a makeshift "mid" scheme, implemented [here] (https://github.com/readium/ts-toolkit/blob/navigator-html/navigator-html/src/comms/mid.ts). The message IDs are useful for advanced debugging, but also *essential* for the Navigator to associate received `_ack` events acknowledging the handling of the event with the original sent command. Both the `navigator-html` and `navigator` packages wrap this system hiding this from developers for convenience. In a module, you simply call the `ack` callback with true/false, and on the web Navigator side, the `send` function allows for optional providing of a callback that will be invoked when the `_ack` is received. Example:

```ts
// Web Navigator
frame.send("go_prev", undefined, ok => {
    console.log("Did we go back a page?", ok);
});
```
```ts
// IFrame (navigator-html)
comms.register("go_prev", ColumnSnapper.moduleName, (_, ack) => {
    const result = doTheWork(); // Result is false if we weren't able to go to previous
    ack(result);
});
```

### Modules

The module class is pretty simple to implement:
```ts
export abstract class Module {
    static readonly moduleName: string;
    abstract mount(wnd: Window, comms: Comms): boolean;
    abstract unmount(wnd: Window, comms: Comms): boolean;
}
```

You just need a unique module name (snake_case), and to implement a mount and unmount function.

The `mount` function is called by the Loader, where modules are mounted one after another, and provided the IFrame window and the `Comms` instance. The Module can then register any commands it should listen to using `comms.register`, and send any commands using `comms.send`. Modules must be designed to not use `comms.send` before the communication handshake has finished (so not during the mount function). The Module can, however, use the `comms.log` function at any time to send logs for debugging purposes, equivalent to using `console.log` on the web.

The `unmount` function should *completely* clean up any event listeners or active modification of the HTML content, and cease all activity. This is because an implementer should be able to mount and unmount Modules at any time, to, for example, switch reading modes. There should be no difference in behavior after mounting and unmounting a Module many times.

Implementers wishing to add their own modules can do so. Exact/official methods for doing so are not set in stone at the moment, but but you *can* add a module to `ModuleLibrary` JS Map, then load and unload it at will by name.

Note that not all basic modules have been implemented yet.

## navigator package

This work is inspired by the [kotlin toolkit](https://github.com/readium/kotlin-toolkit/tree/develop/readium/navigator/), and at the moment implements the reflowable EPUB navigator only. As previously mentioned, the EPUB navigator uses the `navigator-html`, and implements communication with the IFrame. Because this is being done on a web platform, and the IFrames are blobs which are fully accessible by parent frames, the `navigator-html` is actually being executed in the parent, *not* in the IFrame. This makes it much easier as we don't have to inject a build of the `navigator-html` into the IFrame, and the relevant JS is already loaded and warmed (in the JIT/cache) before we even create the IFrame. It might seem somewhat strange that the IFrame helper scripts are being run outside of the IFrame, but this possibility is intentional by design for flexibility. If you're trying to communicate with an IFrame cross-domain (e.g. loading by URL instead of using blobs, for whatever reason) then you would of course, in your custom web Navigator implementation, need to inject the `navigator-html` into the frame on the server-side for security reasons. And then you could still communicate with the IFrame.

On a related note, at the moment, the Navigator still does access the IFrame window directly for certain functionality, but the goal is to completely remove this at a later time so that the navigator can theoretically support interaction/modification of the IFrame purely through the `Comms` message communication. In addition, the current IFrame's window is currently exposed to the implementor through a getter (`EpubNavigator.cframe`) as a temporary (but probably for quite a while temporary until we work out the full scope of the necessary IFrame communication) solution for implementers to meddle with the IFrame directly. Try not to use it where possible.

Note: the EpubNavigator is *not* complete, that's being worked on.

### Frame pool

This feature should be transparent to the implementer, and is intended to greatly enhance the user experience. As the name implies, instead of just a single IFrame being used to load reflowable content, there is a pool of (hidden) IFrames that are preloaded as the user browses the publication. When the user reaches the end of an element, the next element in the readingOrder can be shown nearly instantaneously. This is done by
1. Storing resource content (with injections) as blobs in-memory for the entire duration of the reading session. This means that any re-visitation of a resource requires no additional retrieval request (on the web, an HTTP fetch) 
2. Preloading a limited amount of resources before and after the user's current position (as in Readium locator position) in the publication.

As the user goes fowards, backwards, jumps etc. in the publication, the `FramePoolManager` determines what the current IFrame is, as well what frames to load and dispose of. I consider the use of the position list for this to be by far the "smartest" choice. A more rudimentary piece of logic could, for example, just use the `readingOrder`, and say "load the next 2 resources in the reading order before and after the current one". But this would result in much more inefficient preloading. If there is, for example, a very long chapter, why preload the next resource when it will take the user 50 pages to get to it anyway? Why not instead load it only 5 pages before? That is exactly what the usage of the position list enables, so even as a user navigates through pages in a resource.
I tried to make a diagram illustrating the sliding disposal/creation window, but it didn't turn out that well. I'll leave it here though, in case someone finds it useful:
Illustration of the sliding disposal/creation window
It's probably best just to look at [implementation itself](https://github.com/readium/ts-toolkit/blob/navigator-html/navigator/src/epub/frame/FramePoolManager.ts) though, it's not very long.

Note that right now, the readium CSS that is injected into the IFrame is hardcoded to a jsdelivr URL. This will be fixed to use a local version soon.

Also note that right now, because presentation settings are not yet a part of the package, they cannot be sent before the iframe is shown to the user, so despite the preloading of the IFrame, implementations that use `EpubNavigator.cframe` to inject presentations settings into the iframe won't be able to do it fast enough to prevent an FOUC (flash of unstyled content), especially noticeable with dark mode theme switching.

# Known issues

- If you navigator too quickly through a publication (for example, hold down arrow key), excessive iframes will accumulate in the container element's DOM. Deferring fix to after this PR.
- ColumnSnapper swipes have no velocity


navigator-html/src/Loader.ts
```diff
@@ -0,0 +1,98 @@
+import { CommsReceiveKey, CommsMessage, COMMS_VERSION } from "./comms";
```
mickael-menu: Following our last call, this module could be named `navigator-html-injectables` or `navigator-html-scripts`

navigator-html/src/comms/comms.ts
```diff
@@ -0,0 +1,90 @@
...
+ * adds structure to the messages and lets modules register callbacks.
+ */
+export class Comms {
+    private destination: MessageEventSource | null = null;
+    private registrar = new Map<CommsReceiveKey, Registrant[]>();
+
+    constructor(wnd: Window) {
+        wnd.addEventListener("message", (event) => {
+            if(event.source === null) throw Error("Event source is null");
+            if(typeof event.data !== "object") return; // Skip events whose data is not an object
```
mickael-menu:
```suggestion
            if(typeof event.data !== "object") return;
```

navigator-html/src/modules/setup/Setup.ts
```diff
@@ -0,0 +1,46 @@
...
+        })
+        comms.register("remove_property", "setup", (data: unknown) => {
+            removeProperty(wnd, data as string);
+        })
+
+        return true;
+    }
+
+    unmount(wnd: Window, comms: Comms): boolean {
+        comms.unregister("set_property", "setup");
```
mickael-menu: Maybe this could be simplified and made safer by automatically unregistering all the events of a given module in the `Loader.removeModule()`?

navigator-html/src/comms/keys.ts
```diff
@@ -0,0 +1,25 @@
+
```
mickael-menu: This is basically an implementation of the [Command pattern](https://en.wikipedia.org/wiki/Command_pattern), so to improve clarity I suggest renaming:

* `Send` -> `Event`
* `Receive` -> `Command`

navigator-html/src/comms/comms.ts
```diff
@@ -0,0 +1,90 @@
...
+            }]);
+    }
+
+    public unregister(key: CommsReceiveKey, module: string) {
+        const listeners = this.registrar.get(key);
+        if(!listeners || listeners.length === 0) return;
+        listeners.splice(listeners.findIndex(l => l.module === module), 1);
+    }
+
+    public send(key: CommsSendKey, data: unknown, strict: boolean = false, transfer: Transferable[] = []) {
```
mickael-menu: We will need a way to identify a couple command / event when interfacing with the apps.

For example, when calling `go_id()`, we need to know if it was successful or not.

I suggest adding an optional `id` parameter that the app will need to provide to know when an event is related to the command.

navigator-html/src/comms/keys.ts
```diff
@@ -0,0 +1,25 @@
...
+    "_ping" |
+    "go_prev" |
+    "go_next" |
+    "go_id" |
+    "go_text" |
+    "go_end" |
+    "go_start" |
+    "go_position" |
+    "set_property" |
+    "remove_property";
```
mickael-menu: It looks like it's very specific to each module. Maybe the module themselves should declare what commands they support. Not sure it would work as a type enum in this case.

navigator-html/src/modules/ReflowablePeripherals.ts
```diff
@@ -0,0 +1,54 @@
...
+            x: event.clientX * pixelRatio,
+            y: event.clientY * pixelRatio,
+            targetElement: (event.target as Element).outerHTML,
+            interactiveElement: nearestInteractiveElement(event.target as Element)?.outerHTML
+        }
+
+        // if(handleDecorationClickEvent) TODO
+        //     return;
+
+        this.comms.send("click", clickEvent);
```
mickael-menu: This will be useful for FXL as well, and maybe other media types, so we can have an `Interaction` module handling drags, taps, clicks, etc.

(We recently needed to add drag interaction support in the Kotlin toolkit, this was used in Aldiko to change the device brightness when dragging vertically on the left edge of the screen: https://github.com/readium/kotlin-toolkit/pull/106/files)

navigator-html/src/modules/setup/Setup.ts
```diff
@@ -0,0 +1,46 @@
...
+            false
+        );
+
+        comms.register("set_property", "setup", (data: unknown) => {
+            const kv = data as string[];
+            setProperty(wnd, kv[0], kv[1]);
+        })
+        comms.register("remove_property", "setup", (data: unknown) => {
+            removeProperty(wnd, data as string);
+        })
```
mickael-menu: These two commands will be used only for reflowable as far as I know. They are related to Readium CSS.

navigator-html/src/modules/snapper/ColumnSnapper.ts
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        window.addEventListener("orientationchange", () => { // TODO implement unregister!!!
+            this.snapCurrentOffset();
+        });
+        window.addEventListener("resize", () => { // TODO implement unregister!!!
+            this.snapCurrentOffset();
+        });
+
+        comms.register("go_position", ColumnSnapper.moduleName, (data: unknown) => {
```
mickael-menu: In the legacy glue JS it was called position, but it's ambiguous with the `Locator`'s `position` now.

```suggestion
        comms.register("go_progression", ColumnSnapper.moduleName, (data: unknown) => {
```

navigator-html/src/Loader.ts
```diff
@@ -0,0 +1,98 @@
...
+import { Comms } from "./comms/comms";
+import { Module, ModuleDerived, ModuleLibrary, ModuleName } from "./modules";
+
+/**
+ * The Module loader. Not really sure what else to say right now
+ */
+export class Loader {
```
mickael-menu: We need the `Loader` to be generic over `ModuleName` to allow user to add their own modules.

```suggestion
export class Loader<T> {
```

navigator-html/src/modules/snapper/ColumnSnapper.ts
```diff
@@ -0,0 +1,247 @@
...
+            const position = data as number;
+            if (position < 0 || position > 1) {
+                comms.send("error", {
+                    message: "go_position must be given a position from 0.0 to 1.0"
+                });
+                return;
+            }
```
mickael-menu: We need to make sure that every command calls a callback so that the native/app side doesn't hang waiting for an command to finish. Even if we return an error, it could be represented as a `Result<x>` that the native/app side handles.

navigator-html-injectables/src/Loader.ts
```diff
@@ -0,0 +1,84 @@
+import { CommsCommandKey, CommsMessage, COMMS_VERSION } from "./comms";
+import { Comms } from "./comms/comms";
+import mid from "./comms/mid";
+import { Module, ModuleDerived, ModuleLibrary, ModuleName } from "./modules";
+
+/**
+ * The Module loader. Not really sure what else to say right now
```
mickael-menu: You could mention the object's responsibilities. From what I understand, this one is the entry-point for the `navigator-html-injectables` and manages (mount/unmount) the modules for a single Window object.

navigator-html-injectables/src/Loader.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        if(typeof wnd === 'undefined') // Detect accidental Node/SSR usage
+            throw Error("Loader is not in a web browser");
+
+        if(wnd.parent !== wnd) this.comms.log("Loader is probably in a frame");
+
+        this.loadedModules = uniqueModules.map(name => {
+            const m = ModuleLibrary.get(name); // Find a module with this name
+            if(m === undefined) {
+                console.warn(`Module "${name} does not exist in the library"`)
```
mickael-menu: Shouldn't this warning be sent with `comms.log`?

navigator-html-injectables/src/Loader.ts
```diff
@@ -0,0 +1,84 @@
...
+     * @returns Success
+     */
+    public addModule(moduleName: T): boolean {
+        const m = ModuleLibrary.get(moduleName); // Find a module with this name
+        if(m === undefined) {
+            console.warn(`Module "${moduleName} does not exist in the library"`)
+            return false;
+        }
+        const nm = new m(); // Construct module
+        if(!nm.mount(this.wnd, this.comms)) return false; // Mount module
```
mickael-menu: I think there's an opportunity to refactor this into a (for example) `loadModule(name)` function, as it's duplicated from l31:37.

navigator/README.MD
```diff
@@ -0,0 +1,3 @@
+# navigator
+
+A navigator for web platforms based on the readium Navigator spec. Work is primarily based on the [kotlin toolkit](https://github.com/readium/kotlin-toolkit/tree/develop/readium/navigator/src/main/java/org/readium/r2/navigator), but adapter for the needs of the web. At the moment, this only includes the Reflowable EPUB navigator.
```
mickael-menu:
```suggestion
A navigator for web platforms based on the readium Navigator spec. Work is primarily based on the [kotlin toolkit](https://github.com/readium/kotlin-toolkit/tree/develop/readium/navigator/src/main/java/org/readium/r2/navigator), but adapted for the needs of the web. At the moment, this only includes the Reflowable EPUB navigator.
```

navigator-html-injectables/src/comms/keys.ts
```diff
@@ -0,0 +1,30 @@
...
+    "go_prev" |
+    "go_next" |
+    "go_id" |
+    "go_text" |
+    "go_end" |
+    "go_start" |
+    "go_progression" |
+    "set_property" |
+    "remove_property" |
+    "exact_progress";
```
mickael-menu: It looks like it's very specific to each module. Maybe the module themselves should declare what commands they support? Not sure it would work as a type enum in this case.

Same for the events above.

navigator-html-injectables/src/comms/comms.ts
```diff
@@ -0,0 +1,141 @@
...
+                    this.preLog.forEach(d => this.send("log", d));
+                    this.preLog = [];
+                }
+                return;
+            } else if(this.channelId) {
+                // Enforce matching channel ID and origin
+                if(
+                    data._channel !== this.channelId ||
+                    event.origin !== this.origin
+                ) return;
```
mickael-menu: Do you think it would be worth sending a warning log in this case? Does it give a hint that the Navigator or app is broken?

navigator-html-injectables/src/modules/Module.ts
```diff
@@ -0,0 +1,9 @@
+import { Comms } from "../comms";
+
+export abstract class Module {
+    static readonly moduleName: string;
+    abstract mount(wnd: Window, comms: Comms): boolean;
+    abstract unmount(wnd: Window, comms: Comms): boolean;
+}
```
mickael-menu: Could you fill in the doc comments for these APIs? Maybe using what you wrote in the PR description, in particular:

> The unmount function should completely clean up any event listeners or active modification of the HTML content, and cease all activity. This is because an implementer should be able to mount and unmount Modules at any time, to, for example, switch reading modes. There should be no difference in behavior after mounting and unmounting a Module many times.

navigator-html-injectables/src/modules/ReflowablePeripherals.ts
```diff
@@ -0,0 +1,56 @@
...
+            return;
+
+        const pixelRatio = this.wnd.devicePixelRatio;
+        const clickEvent: FrameClickEvent = {
+            defaultPrevented: event.defaultPrevented,
+            x: event.clientX * pixelRatio,
+            y: event.clientY * pixelRatio,
+            targetElement: (event.target as Element).outerHTML,
+            interactiveElement: nearestInteractiveElement(event.target as Element)?.outerHTML
+        }
```
mickael-menu: Comment for a future PR:

This will be useful for FXL as well, and maybe other media types, so we can have an `Interaction` module handling drags, taps, clicks, etc.

(We recently needed to add drag interaction support in the Kotlin toolkit, this was used in Aldiko to change the device brightness when dragging vertically on the left edge of the screen: https://github.com/readium/kotlin-toolkit/pull/106/files)

navigator/src/Navigator.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Moves to the position in the publication corresponding to the given {Locator}.
+     */
+    abstract go(locator: Locator, animated: boolean, cb: cbv): void;
+
+    /**
+     * Moves to the position in the publication targeted by the given link.
+     */
+    abstract goLink(link: Link, animated: boolean, cb: cbv): void;
```
mickael-menu: These APIs could also fail if, for example, we load a resource whose `href` is not referenced in the manifest.

```suggestion
    abstract go(locator: Locator, animated: boolean, cb: cbb): void;

    /**
     * Moves to the position in the publication targeted by the given link.
     */
    abstract goLink(link: Link, animated: boolean, cb: cbb): void;
```

navigator/src/Navigator.ts
```diff
@@ -0,0 +1,67 @@
...
+        if(this.readingProgression === ReadingProgression.ltr || this.readingProgression === ReadingProgression.ttb || this.readingProgression === ReadingProgression.auto)
+            this.goBackward(animated, completion);
+        else if(this.readingProgression === ReadingProgression.rtl || this.readingProgression === ReadingProgression.btt)
+            this.goForward(animated, completion);
+     }
+
+     /**
+      * Moves to the right content portion (eg. page) relative to the reading progression direction.
+      */
+     goRight(animated = false, completion: cbv) {
```
mickael-menu: The app might want to fallback on showing the menus when reaching the beginning/end of the publication, in which case the boolean is useful in the callback.

```suggestion
     goLeft(animated = false, completion: cbb) {
        if(this.readingProgression === ReadingProgression.ltr || this.readingProgression === ReadingProgression.ttb || this.readingProgression === ReadingProgression.auto)
            this.goBackward(animated, completion);
        else if(this.readingProgression === ReadingProgression.rtl || this.readingProgression === ReadingProgression.btt)
            this.goForward(animated, completion);
     }

     /**
      * Moves to the right content portion (eg. page) relative to the reading progression direction.
      */
     goRight(animated = false, completion: cbb) {
```

navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
```diff
@@ -0,0 +1,323 @@
...
+enum ScrollTouchState {
+    END = 0,
+    START = 1,
+    MOVE = 2
+}
+
+/**
+ * A {Snapper} for reflowable resources using a column-based layout
+ */
+export class ColumnSnapper extends Snapper {
```
mickael-menu, line 19: Some of the `go` commands should be in the `JumpToLocation` module, such as `go_text` and `go_id`. `Snapper`'s job is only to scroll+snap the content when requested.

This implies that some modules like `JumpToLocation` might depend on APIs provided by the `Snapper`. Maybe we can use commands/events for this instead of direct access.

It could flow like this:

1. The app calls a `go_x` command in the `JumpToLocation` module.
2. The command computes a target `DOMRect`, then requests the `Snapper` to `snap_to` the target rect (with a command event, or direct function call).
3. The `Snapper` will resolve the target rect to an actual offset, depending on the technique used for the positioning (CSS translation, native scroll, etc.) and the layout nature (paginated, scrolled, RTL, etc.).
4. The `Snapper` will move the viewport to the target offset, with an animation if requested.

navigator/src/epub/frame/FramePoolManager.ts
```diff
@@ -0,0 +1,136 @@
+import { ModuleName } from "@readium/navigator-html-injectables/src";
+import { Locator, Publication, MediaType } from "@readium/shared/src";
+import FrameManager from "./FrameManager";
+
+const UPPER_BOUNDARY = 5;
+const LOWER_BOUNDARY = 3;
```
mickael-menu, line 7: This is fine for now, but it's the kind of stuff an app might want to change. For example with a `Configuration` object passed to the constructor of `EPUBNavigator`.

navigator/src/epub/frame/FramePoolManager.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export default class FramePoolManager {
```
mickael-menu, line 9: I'm a bit confused about the responsibilities for this class. The name tells me that it holds (and helps to reuse) the loaded iframes, but looking at the implementation it also injects Readium CSS and creates elements for bitmap based resources.

Maybe we could improve clarity by delegating the creation of the iframes to individual function factories? For example:

```javascript
function createHtmlPage(link) -> string? {
    if (!link.mediaType.isHtml()) {
        return null;
    }

    ... inject Readium CSS and scripts
}

function createImagePage(link) -> string? {
    if (!link.mediaType.isBitmap()) {
        return nulll;
    ...
}

// When creating a page, the page factories will be tried in order until one returns non-null content.
FramePoolManager(pageFactories: [createHtmlPage, createImagePage])
```

navigator/src/epub/EpubNavigator.ts
```diff
@@ -0,0 +1,260 @@
...
+        return ReadingProgression.auto; // TODO
+    }
+
+    get publication(): Publication {
+        return this.pub;
+    }
+
+    go(locator: Locator, animated: boolean, cb: () => void): void {
+        const href = locator.href.split("#")[0];
+        // TODO do we let clients go to a resource too? Seems to be the case in kotlin
```
mickael-menu: That's the whole non-linear resources debate 😄 
https://github.com/readium/architecture/discussions/143

For now it's best to prevent it because we need to support it properly to avoid blocking the user or breaking the pagination.

navigator/src/epub/EpubNavigator.ts
```diff
@@ -0,0 +1,260 @@
...
+            this.positions = await this.pub.positionsFromManifest();
+        this.framePool = new FramePoolManager(this.container, this.positions);
+        this.currentLocation = this.positions[0];
+        await this.apply();
+    }
+
+    /**
+     * Exposed to the public to compensate for lack of implemented readium conveniences
+     */
+    public get cframe() {
```
mickael-menu: Maybe we could prefix with an underscore to make it clear that this is an internal API that might be removed in the future?

navigator/src/epub/EpubNavigator.ts
```diff
@@ -0,0 +1,260 @@
...
+export class EpubNavigator extends VisualNavigator {
+    private readonly pub: Publication;
+    private readonly container: HTMLElement;
+    private readonly listeners: EpubNavigatorListeners;
+    private framePool: FramePoolManager;
+    private positions: Locator[];
+    private currentLocation: Locator;
+    private wentBack: boolean;
+
+    constructor(container: HTMLElement, pub: Publication, listeners: EpubNavigatorListeners, positions: Locator[] = []) {
```
mickael-menu: How is the last reading location restored in the reader? On mobile we usually have an `initialLocator` passed to the navigator constructor.

navigator/src/epub/EpubNavigator.ts
```diff
@@ -0,0 +1,260 @@
...
+                            }
+                        }
+                    } else console.log("Clicked on", element);
+                } else {
+                    const oneQuarter = (this.cframe.window.innerWidth * window.devicePixelRatio) / 4;
+                    // open UI if middle screen is clicked/tapped
+                    if (edata.x >= oneQuarter && edata.x <= oneQuarter * 3) this.listeners.click(edata);
+                    // if (scrollMode.value) return; TODO!
+                    if (edata.x < oneQuarter) this.goLeft(false, () => { }); // Go left if left quarter clicked
+                    else if (edata.x > oneQuarter * 3) this.goRight(false, () => { }); // Go right if right quarter clicked
```
mickael-menu: Edge taps to turn pages should be handled in the (test) app. The navigator will just send the tap event.

Each app might want different edge threshold, or even disable this feature for accessibility reasons.

navigator/src/epub/EpubNavigator.ts
```diff
@@ -0,0 +1,260 @@
...
+import { CommsEventKey, ModuleLibrary, ModuleName } from "@readium/navigator-html-injectables/src";
+import { FrameClickEvent } from "@readium/navigator-html-injectables/src/modules/ReflowablePeripherals";
+import * as path from "path-browserify";
+
+export interface EpubNavigatorListeners {
+    frameLoaded: (wnd: Window) => void;
+    positionChanged: (locator: Locator) => void;
+    click: (e: FrameClickEvent) => void;
+    showToc: () => void;
+}
```
mickael-menu: Some of these APIs will need to be moved to a shared `NavigatorListener` interface. But this can be addressed in another PR.

navigator-html-injectables/src/modules/snapper/Snapper.ts
```diff
@@ -0,0 +1,20 @@
+import { Comms } from "../../comms";
+import { Module } from "../Module";
+
+export abstract class Snapper extends Module {
+    mount(wnd: Window, comms: Comms): boolean {
+        const bstyle = wnd.document.body.style;
+        bstyle.userSelect = "none";
```
mickael-menu: This will probably conflict with the (future) highlighting feature which requires being able to select the text.

```suggestion
```

mickael-menu (changes requested):

navigator/src/epub/EpubNavigator.ts
```diff
@@ -264,8 +260,8 @@ export class EpubNavigator extends VisualNavigator {
         }
 
         this.currentLocation = this.positions.find(p => p.href === link!.href)!;
-        const hasProgression = !isNaN(locator.locations.progression as number) && locator.locations.progression! > 0;
-        this.apply(!hasProgression).then(() => {
+        this.apply().then(() => cb());
```
dysbulic:
```suggestion
        this.apply().then(cb);
```

navigator/src/epub/frame/FrameManager.ts
```diff
@@ -68,7 +68,7 @@ export default class FrameManager {
             this.comms?.halt();
     }
 
-    async show(): Promise<void> {
+    async show(atProgress: number | undefined = undefined): Promise<void> {
```
dysbulic:
```suggestion
    async show(atProgress?: number): Promise<void> {
```

navigator/src/epub/frame/FrameManager.ts
```diff
@@ -77,10 +77,19 @@ export default class FrameManager {
         else this.comms = new FrameComms(this.frame.contentWindow!, this.source);
         return new Promise((res, _) => {
             this.comms?.send("focus", undefined, () => {
-                this.frame.style.removeProperty("visibility");
-                this.frame.style.removeProperty("opacity");
-                this.frame.style.removeProperty("pointer-events");
-                res();
+                if(atProgress && atProgress > 0) {
```
dysbulic: `undefined > 0 === false`, so if the compiler will allow it, this could just be:

 ```suggestion
                if(atProgress > 0) {
```

File: navigator/src/epub/frame/FrameManager.ts [hunk 34]
```diff
@@ -77,10 +77,19 @@ export default class FrameManager {
...
+                        this.frame.style.removeProperty("opacity");
+                        this.frame.style.removeProperty("pointer-events");
+                        res();
+                    });
+                } else {
+                    this.frame.style.removeProperty("visibility");
+                    this.frame.style.removeProperty("opacity");
+                    this.frame.style.removeProperty("pointer-events");
+                    res();
+                }
```
dysbulic:
```suggestion
                const remove = () => {
                  this.frame.style.removeProperty("visibility");
                  this.frame.style.removeProperty("opacity");
                  this.frame.style.removeProperty("pointer-events");
                  res();
                }
                if(atProgress && atProgress > 0) {
                    this.comms?.send("go_progression", atProgress, remove);
                } else {
                  remove()
                }
```

navigator/src/epub/frame/FramePoolManager.ts
```diff
@@ -163,8 +163,13 @@ export default class FramePoolManager {
             const newFrame = this.pool.get(newHref)!;
             if(newFrame?.source !== this._currentFrame?.source) {
                 await newFrame.load(modules); // In order to ensure modules match the latest configuration
-                await newFrame.show(); // Show/activate new frame
-                await this._currentFrame?.hide(); // Hide current frame. It's possible it no longer even exists
+
+                await this._currentFrame?.hide(); // Hide current frame. It's possible it no longer even existsin the DOM at this point
+
+                // Update progression if necessary and show the new frame
+                const hasProgression = !isNaN(locator.locations.progression as number) && locator.locations.progression! > 0;
```
dysbulic:
```suggestion
                const hasProgression = locator?.locations?.progression > 0;
```

navigator/src/epub/frame/FrameManager.ts
chocolatkey: 👍

navigator/src/epub/frame/FrameManager.ts
chocolatkey: image
The compiler doesn't like it, I'll leave it for now

navigator/src/epub/frame/FrameManager.ts
chocolatkey: 👍

navigator/src/epub/frame/FramePoolManager.ts
chocolatkey: Still could be undefined, but I'm adding `?? 0` to fix that

navigator-html-injectables/src/Loader.ts
chocolatkey: 👍

navigator-html-injectables/src/Loader.ts
chocolatkey: 👍

navigator-html-injectables/src/modules/Module.ts
chocolatkey: 👍

navigator/src/Navigator.ts
chocolatkey: 👍

navigator/src/Navigator.ts
chocolatkey: 👍

navigator/src/epub/EpubNavigator.ts
chocolatkey: Good idea

navigator-html-injectables/src/comms/keys.ts
chocolatkey: I'll work on this in a future PR

navigator/src/epub/frame/FramePoolManager.ts
chocolatkey: I've added a FrameBlobBuider class for now. I'll use the factory approach when we have more types

navigator/src/epub/EpubNavigator.ts
chocolatkey: I'll handle this in an independent PR as it will break existing consumers of the toolkit

navigator/src/epub/frame/FramePoolManager.ts
chocolatkey: 👍

navigator-html-injectables/src/comms/keys.ts
mickael-menu: Alright

navigator/src/epub/frame/FrameBlobBuilder.ts
```diff
@@ -0,0 +1,85 @@
+import { Link, MediaType, Publication } from "@readium/shared/src";
+
+// TODO embed locally
+const READIUM_CSS_PATH = "https://cdn.jsdelivr.net/gh/readium/readium-css@583011453612e6f695056ab6c086a2c4f4cac9c0/css/dist/{FILE}";
```
mickael-menu, line 4: FYI eventually apps should be able to provide their own build of Readium CSS, as it is customizable at build time.

navigator/src/epub/EpubNavigator.ts
```diff
@@ -108,43 +118,46 @@ export class EpubNavigator extends VisualNavigator {
...
 
     private determineModules() {
         let modules = Array.from(ModuleLibrary.keys()) as ModuleName[];
-        if (this.readingProgression === ReadingProgression.ttb) modules = modules.filter((m) => m !== "column_snapper");
+
+        // Horizontal vs. Vertical reading
+        if (this.readingProgression === ReadingProgression.ttb || this.readingProgression === ReadingProgression.btt)
+            modules = modules.filter((m) => m !== "column_snapper");
+        else
+            modules = modules.filter((m) => m !== "scroll_snapper");
```
mickael-menu, line 136: It will be relevant for a future PR, but FYI:
* reading progression will have only for possible values LTR and RTL
* to determine whether we scroll or paginate, there will be a setting `scroll: bool`
* we'll have an additional `scrollAxis: horizontal | vertical` setting to support CJK in horizontal scroll mode

navigator-html-injectables/src/modules/ReflowablePeripherals.ts
```diff
@@ -15,39 +15,69 @@ export class ReflowablePeripherals extends Module {
...
             // There's an ongoing selection, the tap will dismiss it so we don't forward it.
             return;
 
+        // if(handleDecorationClickEvent) TODO
+        //     return;
+        if(!event.isPrimary) return;
+
         const pixelRatio = this.wnd.devicePixelRatio;
-        const clickEvent: FrameClickEvent = {
+        event.preventDefault(); // TODO when not to prevent?
```
mickael-menu, line 36: Are you able to get a bool result from the tap/click events? Then you can delegate this choice to the app.

See https://github.com/readium/kotlin-toolkit/blob/e5231436cac3a802fb5e406eda63bce4f1a5549d/readium/navigator/src/main/assets/_scripts/src/gestures.js#L33-L40

mickael-menu (approved):
There are some pending comments that you can review, but nothing preventing this PR from being merged.

navigator/src/epub/frame/FrameBlobBuilder.ts
chocolatkey: Yeah, in my local copy I have readium CSS as a submodule, I'm still deciding on the best way to integrate it. This is just a temporary solution, definitely shouldn't be including like this in the final vers, e.g. prevents offline reading.

Continuing here: https://github.com/readium/ts-toolkit/issues/18

navigator/src/epub/EpubNavigator.ts
chocolatkey: OK, @mickael-menu can you link me to where those changes were discussed for other toolkits/spec? Also, let's continue here https://github.com/readium/ts-toolkit/issues/19

navigator-html-injectables/src/modules/ReflowablePeripherals.ts
chocolatkey: That can't be done unfortunately because you can't make the browser wait for completion of an async function **before** calling preventDefault(). Since the injectables communicate with an async message to send the event, and then another back that would contain the bool you speak of, the event will have already happened and will no longer be preventable. I'm still thinking of a solution, suggestions welcome. Maybe I'm missing something obvious.

Continuing here: https://github.com/readium/ts-toolkit/issues/20

PR #32, opened 2024-03-07
Deselection of text when moving/changing resources

chocolatkey:
As discussed in today's Readium call, this implements deselection of any selection in the iframes of the FXL and reflowable EPUBs when it is aesthetically unpleasing to keep them displaying, such as when swiping to a previous/next resource in the horizontal reader.
This feature is already implemented to some extent in the mobile toolkits.


mickael-menu (approved):
Thanks @chocolatkey 👌

mickael-menu:
I'll let you resolve the conflict and then feel free to merge.

PR #46, opened 2024-08-12
Guided navigation

chocolatkey:
Adds support for parsing of Guided Navigation Documents, as well as the AudioEngine interface and WebAudioEngine class


chocolatkey:
@JayPanoz do you think we should merge this before or after your changes?

JayPanoz:
@chocolatkey go ahead and merge it before. That way it will ship with the initial packages and we can double-check everything is OK. 🙏

PR #52, opened 2024-09-17
Fix viewport in ReflowableSetup

JayPanoz:
Not sure I’m missing something – there’s nothing obvious while blaming the file – but the following created an issue when switching to devices with a pixel density that is more than `1` in Readium Playground. 

https://github.com/readium/ts-toolkit/blob/bb3a594f5e7f78fbd4aa444cbf283475c4b34ce4/navigator-html-injectables/src/modules/setup/ReflowableSetup.ts#L11-L15

Switched to an actual Android device to make sure it was not a software issue of sorts and this is what I would get:

Multiple columns instead of 1.

The columns are cut off and not perfectly aligned.

Note I used the desktop version so that it’s a little bit clearer what happens, otherwise columns are like a few characters wide. But I can guarantee the problem applies w/o using the desktop version – as well as software emulation e.g. Safari’s responsive debug, etc.

While this is the expected result:

A single column

It’s working as expected when removing the `devicePixelRatio` but again, maybe I’m missing something cos’ I can’t explain why it was used in the first place so maybe it solves an issue.

### DIFF
--- a/navigator-html-injectables/src/modules/setup/ReflowableSetup.ts
+++ b/navigator-html-injectables/src/modules/setup/ReflowableSetup.ts
@@ -11,7 +11,7 @@ export class ReflowableSetup extends Setup {
     onViewportWidthChanged(event: Event) {
         const wnd = event.target as Window;
         // const pageWidth = wnd.innerWidth / wnd.devicePixelRatio;
-        setProperty(wnd, "--RS__viewportWidth", `calc(${wnd.innerWidth}px / ${wnd.devicePixelRatio})`);
+        setProperty(wnd, "--RS__viewportWidth", `${wnd.innerWidth}px`);
     }
 
     mount(wnd: Window, comms: Comms): boolean {


JayPanoz:
Alright got the reason why that happens. 

It isn’t creating an issue in the existing readers using navigator because the `--RS__viewportWidth` property is actually not used – master branch has a slightly older version of v1 than what was present in develop when I restarted working on ReadiumCSS and it was just set to 100% (see right pane). With up-to-date version of ReadiumCSS, it is taking the `var` into account to size the html element (see left pane), which is why it’s half the size of the `iframe` on `2x` pixel ratio devices, a third on `3x`, etc. 

With the newer version of ReadiumCSS on the left, --RS__viewportWidth is used for min, max, and width. On the right, the older version of ReadiumCSS just set them to 100%

So this change should fix the issue while being transparent, with no side-effect, to existing readers.

PR #55, opened 2024-09-22
Improve loading of resources with many images

chocolatkey:
In a reflowable reader, the document's resources must be fully loaded in order for the positioning to be correct. That means every single image must be fully loaded, so the height and width it takes in the document is determined, and the e.g. columnated layout is fully finished. The ts-toolkit does allow proceeding fowards/backwards to a document until it has fully loaded, so when a resource takes particularly long to load, the user is left waiting, and appears stuck in the reader.
The visual queues for this should be improved in a future version of the reflowable reading engine itself, but in the meantime there is a larger issue that has to be addressed - <iframe> documents with hundreds of images take a *long* time to load when streamed over the web, because, based on my understanding, the loading of the frames' subresources are deprioritized by browser heuristics.
Here is an example of the loading of a reader using the ts-toolkit, in Google Chrome, before this change:
dev tools network requests
As you can see, the priority of the requests is low, and they are performed sequentially.

After this optimization, the images all have high priority, and are loaded in parallel by the browser:
dev tools network requests

While I'm not quite happy with the abuse of the fetchpriority attribute to fix this issue, and the further modification of the original document, I don't think there's any other choice.

### DIFF
--- a/navigator/src/epub/frame/FrameBlobBuilder.ts
+++ b/navigator/src/epub/frame/FrameBlobBuilder.ts
@@ -172,6 +172,17 @@ export default class FrameBlobBuider {
             // Readium CSS After
             doc.head.appendChild(styleify(doc, cached("ReadiumCSS-after", () => blobify(stripCSS(readiumCSSAfter), "text/css"))));
         }
+
+        // Set all <img> elements to high priority
+        // From what I understand, browser heuristics
+        // de-prioritize <iframe> resources. This causes the <img>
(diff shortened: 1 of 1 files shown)


JayPanoz (approved):
Yeah I can see why you’re not particularly happy with this solution but as far as I can remember, link `preload`/`prefetch` can’t really help with this issue and there’s not much else you can use. 

Looks good enough to me at the moment, so that there’s time to design and develop a more thorough solution.

PR #58, opened 2024-10-01
Add CSS selector to tap/click events

chocolatkey:
This PR adds the CSS selector to the `FrameClickEvent` interface, which as proven useful for implementer purposes.
I also standardized the usage of the `ReadiumWindow` type across the navigator-html-injectables package, and renamed `ReflowablePeripherals` to `Peripherals` since it's actually used in FXL EPUBs as well

### DIFF
--- a/navigator-html-injectables/src/Loader.ts
+++ b/navigator-html-injectables/src/Loader.ts
@@ -1,4 +1,5 @@
 import { Comms } from "./comms/comms";
+import { ReadiumWindow } from "./helpers/dom";
 import { Module, ModuleDerived, ModuleLibrary, ModuleName } from "./modules";
 
 /**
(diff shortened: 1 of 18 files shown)


HadrienGardeur:
@chocolatkey do you mind sharing the use case? Was it for a tap on text (definition) or an image (open in a separate view) for example?

chocolatkey:
@HadrienGardeur It's because we're attempting to add support for playback of media overlay EPUBs starting at a specific element, through the user clicking on it.
There is an FXL EPUB with HTML similar to the following:
```html
<div id="p2">
  <span class="something">Content</span>
</div>
```
where p2 is an anchor linked in the guided navigation document. When the user clicks on the text, only the `<span>` and its contents are sent back as the `interactiveElement` (which is correct), and it is therefore very difficult to determine linkage is to the media overlay. There were ideas internally at DeMarque to use the `x` and the `y` position, but that has much more potential for bugs. So I suggested that we should add the CSS selector as another property, because the ID can be extracted out of it by crawling up the selector tree until an anchor element with a matching guided navigation element is found. And it may be useful for other implementers as well trying to locate the element that was clicked in the document.

JayPanoz (approved):
LGTM

PR #59, opened 2024-10-01
Fix authored CSS animations

chocolatkey:
Certain fancy FXL publications from publishers include CSS animations that expect to begin when the user first sees the page. The ts-toolkit, however, preloads the frames of publications in the background, before they are visible to the user. This messes up the animation, because it is likely already finished by the time the user sees it. This PR addresses that by capturing all the animations that are playing when the iframe is loaded (technically, a tiny bit after it's loaded), and then restarting them every time the frame is shown to the user. The animations are also paused when the frame goes out of view.

Example of improved support:

https://github.com/user-attachments/assets/04f24f05-47ac-4c37-b912-ada7c2cf3e0a

### DIFF
--- a/navigator-html-injectables/src/modules/setup/Setup.ts
+++ b/navigator-html-injectables/src/modules/setup/Setup.ts
@@ -7,7 +7,6 @@ export abstract class Setup extends Module {
     static readonly moduleName: ModuleName = "setup";
 
     private comms!: Comms;
-    private wnd!: ReadiumWindow;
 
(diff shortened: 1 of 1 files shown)


chocolatkey:
I just want to add, I think this authoring is a really bad idea, because it's assuming that the media overlay playback also begins when the document is shown, which is not necessarily the case. If it's paused, stopped, restarted, then it's out of sync with the animation, and since the animation shows/hides the text, it's particularly confusing

JayPanoz (approved):

PR #60, opened 2024-10-03
Add Position List to Known MediaTypes

JayPanoz:
Adding the MediaType of the Position List to shared models should help with discovery and usage (auto-completion, etc.)


chocolatkey:
@JayPanoz while you're at it, would you mind adding the rest of the missing mimetypes? Here are the equivalents in Go: https://github.com/readium/go-toolkit/blob/develop/pkg/mediatype/types.go#L40-L48

JayPanoz:
@chocolatkey good idea, will do.

PR #62, opened 2024-10-07
Current numbers range for reflow EPUB

JayPanoz:
Snappers report the progression to EPUBNavigator so that it can find the nearest position, which results in an array with a single Locator position when getting `currentNumbers` in reflow EPUB. In FXL it reports `2` when spreads are displayed.

This PR aims to return the first and last locators’ position to be found in a range by having snappers report a reference progression – a.k.a. progression by a `ReadiumWindow` viewport – in addition to the current one. 

This reference can then be used in EPUBNavigator to build a range, and find the aforementioned positions. If there is no `last` found, then it will return only the first one, as it did previously.

### DIFF
--- a/navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
+++ b/navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
@@ -40,7 +40,7 @@ export class ColumnSnapper extends Snapper {
     }
 
     reportProgress() {
-        this.comms.send("progress", this.wnd.scrollX / this.cachedScrollWidth);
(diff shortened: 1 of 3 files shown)


JayPanoz:
Just to explain why this is still a draft PR, I'm willing to test scroll first before this is merged, as I don't want to introduce new bugs and issues for existing apps using Navigator. 

So bear with me until I properly implement scroll in Readium Playground. 🙏

JayPanoz:
Actually this should be reviewed now and merged if OK as scroll in Playground has progress de-prioritized for the time being so there’s not much incentive to wait until that PR goes stale, especially as progress is not reported on scroll yet.

chocolatkey (approved):
LGTM

PR #64, opened 2024-10-07
Add aria-hidden to hidden iframes in navigator

chocolatkey:


JayPanoz (approved):
LGTM.

In case anyone wonders why this is added, we’re trying to hide hidden iframes specifically to AT since `visibility: hidden` is not enough with some screen readers.

PR #98, opened 2024-12-17
Fix workspace in navigator-html-injectable

JayPanoz:
Otherwise it fails to `pnpm install/add` with error:

```
ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In navigator-html-injectables: No matching version found for @readium/shared@workspace:1.2.0 inside the workspace
```

### DIFF
--- a/navigator-html-injectables/package.json
+++ b/navigator-html-injectables/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "@juggle/resize-observer": "^3.4.0",
-    "@readium/shared": "workspace:1.2.0",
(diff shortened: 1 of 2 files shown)


chocolatkey (commented):
Could you please do the same (asterisk wildcard) for all the workspace dependencies across all the ts-toolkit packages? Thanks

JayPanoz:
That should have been the only one – double-checked this morning – but I updated the lock file so that it’s sync’d.

chocolatkey (approved):

PR #99, opened 2025-01-10
First visible locator (EPUBS) improvements

oscar-rivera-demarque:
We have detected some different publications where the first visible locator is not working as expected. Doing an investigation, most of the cases are related to a crash in the usage of "cssSelectorGenerator" method. Additionally, there is another case covered here, when a publication has a list which is long enough to be shown entirely in the current pages.

Cases explanation:

In the first case, I've detected that, when using TTS for some publications in which the elements have custom properties (eg. xml:lang), the css selector process will crash, so the first visible locator will not be in the actual current page, but in the page in which the latest known element is.
Screenshot 2025-01-10 at 3 04 01 PM
________________________
1a- Current behavior in _Sept comme setteur_ publication: 
Notice that even if I try to start in the page 46, it will go back to the page 40 (the beginning of the current chapter).

https://github.com/user-attachments/assets/203f7bd0-3a13-4bf5-bde9-2118b88b20f1

1b- New behavior in _Sept comme setteur_ publication: 
As now the cssSelectorGenerator is not crashing, it's able to get the right element to start.

https://github.com/user-attachments/assets/ea4a8d70-38bd-441f-870b-50cfc47d11ff

__________________________

In the second case, the publication I found is a little bit special, most of the content is wrapped into an UL element. 
This situation make the whole list of elements to act as list items, and that's why the current `"display: block"` validation is not enough to detect the right element to be selected as the first visible element.
Screenshot 2025-01-10 at 3 09 26 PM

___________________________

2a- Current behavior in _CRAAV Therapy_ publication:
Notice that I'm starting in the page 72, but as most of the content is ignored, the "first visible locator" will be the beginning of the UL (in page 5)

https://github.com/user-attachments/assets/80b40da4-2e56-475e-818f-9babb8114857

2b- New behavior in _CRAAV Therapy_ publication:
As now the elements are not ignored, the first visible locator is actually getting the right one.

https://github.com/user-attachments/assets/facc5880-03cc-4091-af36-e2abfcad9c83

### DIFF
--- a/navigator-html-injectables/src/helpers/dom.ts
+++ b/navigator-html-injectables/src/helpers/dom.ts
@@ -68,13 +68,38 @@ export function nearestInteractiveElement(element: Element): Element | null {
 /// Returns the `Locator` object to the first block element that is visible on
 /// the screen.
 export function findFirstVisibleLocator(wnd: ReadiumWindow, scrolling: boolean) {
-    const element = findElement(wnd, wnd.document.body, scrolling);
+    const element = findElement(wnd, wnd.document.body, scrolling) as HTMLElement;
+    const removedAttributes = new Map<HTMLElement, Attr[]>();
+    if (element) {
+        // remove all the properties other than class, id or style to avoid breaking the cssSelector
+        for (let i = 0; i < element.attributes?.length; i++) {
+            const attr = element.attributes[i];
+            if (attr.name !== "class" && attr.name !== "id" && attr.name !== "style") {
+                if(!removedAttributes.has(element)) {
+                    removedAttributes.set(element, [attr])
+                } else {
+                    removedAttributes.set(element, removedAttributes.get(element)!.concat([attr]))
+                }
+                element.removeAttribute(attr.name);
+            }
+        }
+    }
+    
+    const cssSelector = wnd._readium_cssSelectorGenerator.getCssSelector(element);
+    // Now that the cssSelector is generated, we can put back the removed attributes
+    if (removedAttributes.size > 0) {
+        removedAttributes.forEach((attrs, elem) => {
+            attrs.forEach(attr => {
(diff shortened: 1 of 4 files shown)


JayPanoz (approved):

navigator-html-injectables/src/helpers/dom.ts
```diff
@@ -111,11 +120,29 @@ function isElementVisible(wnd: ReadiumWindow, element: Element, scrolling: boole
...
+}
+
 function shouldIgnoreElement(element: Element) {
     const elStyle = getComputedStyle(element);
     if (elStyle) {
         const display = elStyle.getPropertyValue("display");
-        if (display != "block") {
+        // Added list-item as it is a common display property for list items
+        // TODO: Check if there are other display properties that should be ignored/considered
+        if (display != "block" && display != "list-item") {
```
chocolatkey, line 145: @oscar-rivera-demarque can you explain why we can only allow block and and list-item as the display values? For example, why not inline-block?

chocolatkey (approved):
Looks good, simplifying the CSS selectors is probably a good idea regardless of these bugs for cross-compatibility across reading systems and web engines

navigator-html-injectables/src/helpers/dom.ts
oscar-rivera-demarque: @chocolatkey well, that was actually the question I had about why the display block was necessary here? (As it was working with that validation before). 
I'm not sure what we're validating using the `display != "block"` line. I assumed it's issued because at some point the elements had the display behavior changed based on if they were loaded/allowed, or not. But this time I was not able to confirm if that was the expectation.
That's why I just simply added the display list-item, as a new allowed property, so that way I don't interfere in the current working behavior, but for sure we can review if maybe this check is not even necessary.

PR #106, opened 2025-02-07
Preferences API

JayPanoz:
This branch is meant to implementing the [Preferences API](https://readium.org/architecture/proposals/009-preferences-api.html) in the ts-toolkit Navigator, bringing it on par with – and taking inspiration from – Readium’s Mobile toolkits. 

This will make it a lot easier to implement user settings in apps and is using ReadiumCSS version 2. 

It will be merged as a new major version since it will consequently come with breaking changes.

In ReadiumCSS v1, `auto` pagination was built-in, switching from 1 to 2 columns depending on the window/screen width. This is no longer the case. However, we plan to replace this feature with preferences that can be set:

- an optimal line length, in `ch` (number of characters)
- a minimal line length: can be `ch`, can be `undefined`, or `null`.

The logic is the following:

- the number of columns is no longer limited to `2`. It can be as many as apps see fit
- the number of columns will be computed based on this optimal line length in `auto` mode
- the minimal line length is used to apply a lower bound even if the number of columns is larger than `1`, dropping a column in the process if it can’t fit
- in case this minimal line length is `undefined`, then optimal line-length is the minimal one
- in case it’s `null` then this lower bound is disabled entirely and apps can enforce the number of columns they want

This is currently implemented in [Readium/Playground](https://github.com/readium/playground) and will be ported back into the ts-toolkit.


JayPanoz:
Update: there are some additional changes coming (line-length, removal of advanced settings/publisherStyles) so I have to delay things a little bit.

PR #126, opened 2025-06-17
Improve SSR import-ability of toolkit

chocolatkey:
An attempt was made to use the ts-toolkit (more specifically, thorium-web) in combination with Next.JS SSR. In the SSR context, the global `window` is not available. In the injectables, we check for highlighting support in the browser, which was done by having a global const `const canNativeHighlight = "Highlight" in window;`. This broke the SSR because `window` is undefined. So I've changed it to a function instead so it won't get executed on import.

### DIFF
--- a/navigator-html-injectables/package.json
+++ b/navigator-html-injectables/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@readium/navigator-html-injectables",
-  "version": "2.0.0-beta.2",
+  "version": "2.0.0-beta.3",
   "type": "module",
   "description": "An embeddable solution for connecting frames of HTML publications with a Readium Navigator",
   "author": "readium",
--- a/navigator-html-injectables/src/modules/Decorator.ts
+++ b/navigator-html-injectables/src/modules/Decorator.ts
@@ -49,7 +49,7 @@ interface DecorationItem {
     container: HTMLElement | undefined;
 }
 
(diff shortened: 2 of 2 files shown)


JayPanoz:
LGTM, I could quickly test it and was able to use the `StatefulReader` component in next.js without a `dynamic` import. 

I have taken the liberty to bump the injectables’ package version so that I can publish it immediately so that you do not have to bother about this

JayPanoz (approved):

JayPanoz:
After retry-ing the updated packages in Thorium Web proper (and not a test repo where it did not happen), I can see we also have an error with `navigator` being undefined on reload, [due to the sML helper](https://github.com/readium/ts-toolkit/blob/develop/navigator/src/helpers/sML.ts).

chocolatkey:
@JayPanoz should I go ahead and fix that?

JayPanoz:
Probably, I was not sure how to best deal with it since it's code from another source, and I preferred to extend it for the iPadOSPatch myself so that we don't have code that has potential to go out-of-sync. 

I wanted to discuss this first because of that but I know we will end up using your solution anyway.

chocolatkey:
@JayPanoz OK, it's in the new PR. Regarding the sync of the code, I've checked and there are no modifications in the original source since I last ported it.

PR #127, opened 2025-06-19
Improvements to sML library

chocolatkey:
- Add fallbacks for when `navigator` is not available in the JS globals
- Add typing for `navigator.userAgentData`


JayPanoz (approved):
LGTM. 

Sorry for the delay, the idea was to merge this, then update ReadiumCSS (and Preferences API), and finally publish so that we handle everything in a single version on npm if that’s OK. 

I will then update Thorium Web and publish its package so that everything is up-to-date.

PR #138, opened 2025-07-02
Fix Webkit scroll issue

### LINKED ISSUE #136: Scroll is not scrollable in Webkit when loading the resource directly in scroll
Not a regression but a discovery: in Webkit, scrollable resources cannot be scrolled unless a re-layout is forced after load…

For some reason all the dimensions are correct but WebKit thinks the scrolling element cannot be scrolled. Maybe this is a side effect of columns being involved (as we have to reset them for scroll) I am not sure because...

It is insanely difficult to debug because it will go away on reflow/re-layout. Which means disabling some style on root or body resolves the issue, and it's impossible to pinpoint the exact culprit, nor confirm this is because of columns being involved.

This bug was accidentally resolved by previously injecting the scroll affordances. Checked with an empty `div` and an empty `span` after load, it resolves the issue too.

This is also why it works when switching from paged to scroll in the same resource – in that case, we have to update the custom properties on root. But when it's another resource, we load it directly with these properties.

Also why it works if you change settings...

We will consequently have to force a reflow in scroll for Safari. 

I will probably attempt to handle this in the Scroll snapper. Best case scenario it's just about removing and re-adding an inline style but there are few we can rely upon, primarily ReadiumCSS scroll flag.

Other options include appending a dummy element in the DOM, which would at least be consistent with the virtual columns in the Column Snapper if that can help avoid specificities in future helpers, etc.

JayPanoz:
Resolves #136

I am really open to any alternative idea but it is pretty much the only working resolution I could achieve due to getting around Webkit’s lifecycle issues e.g. it won’t even work if it is not called at the end of `go_progression` when it’s used, and it works only partially in the sense Webkit doesn’t graphically render the overflowing content – with a reference of the viewport to render  `progression` is `0`, hence why the triggering of the scroll event. 

Really tried to make it happen before `show` as well, but there is pretty much nothing you can rely upon if you want to use a resize or mutation observer…

It is worth noting I cannot fully explain what the bug is, based on the fact it is pretty much impossible to debug, a recalc/reflow resolving the non-scrollable issue. 

@chocolatkey Maybe another pair of eyes can spot something I did not consider in the last 2 days. Please let me know if you need clarifications as this bug and its resolution are pretty convoluted.

### DIFF
--- a/navigator-html-injectables/package.json
+++ b/navigator-html-injectables/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@readium/navigator-html-injectables",
-  "version": "2.0.0-beta.3",
+  "version": "2.0.0-beta.4",
   "type": "module",
   "description": "An embeddable solution for connecting frames of HTML publications with a Readium Navigator",
   "author": "readium",
--- a/navigator-html-injectables/src/comms/keys.ts
+++ b/navigator-html-injectables/src/comms/keys.ts
@@ -41,5 +41,5 @@ export type CommsCommandKey =
     "focus" |
     "activate" |
     "shake" |
-    "force_webkit_reflow";
+    "force_webkit_recalc";
 ;
--- a/navigator-html-injectables/src/comms/keys.ts
+++ b/navigator-html-injectables/src/comms/keys.ts
@@ -40,5 +40,6 @@ export type CommsCommandKey =
     "unfocus" |
     "focus" |
     "activate" |
-    "shake";
+    "shake" |
+    "force_webkit_reflow";
 ;
--- a/navigator-html-injectables/src/helpers/document.ts
+++ b/navigator-html-injectables/src/helpers/document.ts
@@ -79,3 +79,19 @@ export function appendVirtualColumnIfNeeded(wnd: ReadiumWindow): boolean {
 
     return virtualColsCount !== needed;
 }
+
+/**
+ * This forces a recalculation in WebKit browsers.
+ * It is needed in scroll mode to ensure that the content is scrollable.
+ * Webkit seems to find itself in some kind of limbo if we do not do that.
+ * It has everything correct but the scroll listener is non-functional, 
+ * unless you force a recalc or reflow.
+ * It is not needed in paginated mode.
+ */
+export function forceWebkitRecalc(wnd: ReadiumWindow) {
+    // Borrowed from APB themselves… 
+    const styleElement = wnd.document.createElement("style");
+    styleElement.appendChild(wnd.document.createTextNode("*{}"));
+    wnd.document.body.appendChild(styleElement);
+    wnd.document.body.removeChild(styleElement);
+}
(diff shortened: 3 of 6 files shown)


chocolatkey:
@JayPanoz Which webkit-based platforms were you trying this out on? OSX, iPhone, iPad, or all?

JayPanoz:
All. Basically had Safari locally on MacOS, iPadOS and iOS on BrowserStack.

My gut feeling tells me it applies to all browsers using WebKit though, due to the fact it smells like a race condition during rendering. I have not checked the source code of WebKit because I have no clue where to start as I can't fully explain this bug but it's nasty.

chocolatkey:
@JayPanoz OK, I'll try a few things just to make sure I can't find anything. Were you using thorium web or the vanilla reader in this repo? I'll probably try using the vanilla reader

JayPanoz:
Thorium Web, I guess it is a good idea to test the vanilla reader to scope down the issue. 

It should be ready-to-reproduce on [scroll-ui.thorium-web.pages.dev](scroll-ui.thorium-web.pages.dev)

Just to make sure, the pattern to reproduce is:

- when switching from paged to scroll it works without any issue – the custom property is set dynamically 
- while in scroll, once you navigate to a resource which was not in the pool, scroll event does not register – the custom property was added into the blob.

While scroll does not register, the scrolling element has no issue scrolling through `scrollTo`, `scrollTop`, etc. 

Any reflow will make the bug disappear e.g. resizing the window. This means the app shell can accidentally resolve the issue on re-render, etc.

chocolatkey:
@JayPanoz FYI, I confirmed the behavior in iBooks:
```js
_forceStyleRecalc = function() {
    const e = document.createElement('style');
    e.appendChild(document.createTextNode('*{}')), document.body.appendChild(e), document.body.removeChild(e)
}
```

chocolatkey:
@JayPanoz I haven't found the culprit yet, and I don't want to hold up your progress longer. As you mentioned, debugging it is very tedious. iOS 26 doesn't make a difference.
Have you considered filing a webkit bug with a minimal reproducible testcase?

JayPanoz:
@chocolatkey No worries, I was focusing on https://github.com/readium/css/pull/190 since I have a gut feeling they may be related in some way (rendering path on scroll) and one issue could help explain the other. 

I tried removing components to make sure they were not interacting so there was literally the container left in the app, tried removing CSS that could be problematic from various SO threads in the app, then in Readium CSS, etc. But I have zero clue at the moment. 

This is the fundamental problem with both issues: I can’t really tell why they happen, and it’s nearly impossible to create a minimal reproductible test case because of that.

JayPanoz:
I will merge this fix as it’s blocking for Thorium Web v.1 but will continue debugging in order to find a nicer alternative in the upcoming weeks, and report the issue in Webkit tracker. Hopefully it should be trivial reverting if we can implement something else, or resolve the issue at its root so that it does not happen in the first place.

PR #139, opened 2025-07-02
Enable progression 0 on frame show

### LINKED ISSUE #132: Go method may not work as expected with frames still in the pool
This became quite obvious when testing the new scroll-ui of Thorium Web but it looks like when `progression` is `0` in the `Locator` and the frame is still in the pool in reflow, when navigating to the resource, its `progression` will not be reset. 

I also noticed a similar issue in paginated when using the jump to position, which is using the same `go` method.

JayPanoz:
Resolves #132

This is particularly visible when using `go()` with a locator whose `progression` is explicitly `0`, because frames still in the pool will not update their progression as expected in that specific case and show at their previous progression in memory.

That can happen with features such as jump to location/position.

### DIFF
--- a/navigator/src/epub/frame/FrameManager.ts
+++ b/navigator/src/epub/frame/FrameManager.ts
@@ -97,7 +97,7 @@ export class FrameManager {
                         this.hidden = false;
                         res();
                     }
-                    if(atProgress && atProgress > 0) {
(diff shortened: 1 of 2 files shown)


chocolatkey (approved):
LGTM

Issue #66, opened 2024-10-15
positionChanged doesn’t fire reliably for FXL


JayPanoz:
It looks like `positionChanged` event doesn’t necessarily fire for FXL after it did at least once for a spread. It causes a discrepancy between reflow and FXL where your functions are not guaranteed to run for FXL.

Currently Playground is handling the progression and saving the current location through `positionChanged` but, as you can see on [the latest deploy](https://b4dd12ff.playground-7xz.pages.dev/read?book=https%3A%2F%2Fpublication-server.readium.org%2FQmVsbGFPcmlnaW5hbDMuZXB1Yg) the numbers at the bottom of the page may occasionally get stuck, if you go forward multiple times, and then go backward e.g. you’re back on pages `2–3 of 29` but the progression at the bottom is still showing pages `5–6 of 29`.

Correct me if I’m wrong but it’s my understanding FXL doesn’t register a `progress` message that Navigator can use to fire `positionChanged`, unlike the Reflow snappers, so `first_visible_locator` is, unless I’m missing something, the only thing firing `positionChanged` for FXL. 

Navigator’s `currentPositionNumbers` are correct though, so there are ways around this issue if you need to display them e.g. call on show/hide bottom bar, or navigation actions, except for swipe as it is not implemented yet (#63). In other words, if you want to always display these but find yourself swiping on a mobile, these numbers will not be correct. And it can become quite tricky to keep them in sync as you’ll have to rely on some more complex logic to check when the user has progressed.

chocolatkey:
@JayPanoz I'll take a look

HadrienGardeur:
Any progress on that one?

PR #141, opened 2025-07-03
Implement viewport + helpers in EpubNavigator

JayPanoz:
This replaces currentNumbers with the new concept of `viewport`, as already implemented in Swift, and adds helpers that are using `viewport` to infer whether we are:

- at the start of the resource;
- at the end of the resource;
- at the start of the publication;
- at the end of the publication.

`Viewport` contains:

- `readingOrder`: the visible frames
- `progression`: the progression for each item in readingOrder
- `positions`: the positions for each item in `readingOrder`

It is worth noting `ColumnSnapper` and `ScrollSnapper` are now reporting a progression range so that we can infer a value of `1` at the end of the resource. 

This has been implemented in a non-disruptive way, replacing `currentNumbers` using the same logic – FXL is currently implemented as its own method in `FramePoolManager` while reflow is implemented in `EpubNavigator`. 

I tried leveraging the opportunity to handle issue #66 but this is less trivial than anticipated due to discrepancies in how FXL and reflow work, and I did not want to introduce unintended side-effects. I do think this should be its own PR with discussion informing the change.

### DIFF
--- a/navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
+++ b/navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
@@ -39,7 +39,14 @@ export class ColumnSnapper extends Snapper {
     }
 
     reportProgress() {
-        this.comms.send("progress", { progress: this.wnd.scrollX / this.cachedScrollWidth, reference: this.wnd.innerWidth / this.doc().scrollWidth });
+        const scrollX = this.wnd.scrollX;
+        const scrollWidth = this.cachedScrollWidth;
+        const progress = Math.max(0, Math.min(1, scrollX / scrollWidth));
+        const viewportEnd = Math.max(0, Math.min(1, (scrollX + this.wnd.innerWidth) / scrollWidth));
+        this.comms.send("progress", {
+            start: progress,
+            end: viewportEnd
+        });
     }
 
     private shakeTimeout = 0;
--- a/navigator/src/epub/EpubNavigator.ts
+++ b/navigator/src/epub/EpubNavigator.ts
@@ -640,21 +640,23 @@ export class EpubNavigator extends VisualNavigator implements Configurable<Confi
     }
 
     get isResourceStart(): boolean {
-        const currentProgression = this.viewport.progressions.get(this.currentLocation?.href);
-        return currentProgression?.start === 0;
+        const firstHref = this.viewport.readingOrder[0];
+        const progression = this.viewport.progressions.get(firstHref);
+        return progression?.start === 0;
     }
     
     get isResourceEnd(): boolean {
-        const currentProgression = this.viewport.progressions.get(this.currentLocation?.href);
-        return currentProgression?.end === 1;
+        const lastHref = this.viewport.readingOrder[this.viewport.readingOrder.length - 1];
+        const progression = this.viewport.progressions.get(lastHref);
+        return progression?.end === 1;
     }
     
-    get isPublicationStart(): boolean {
+    get canGoBackward(): boolean {
         const firstResource = this.pub.readingOrder.items[0]?.href;
         return this.viewport.progressions.has(firstResource) && this.viewport.progressions.get(firstResource)?.start === 0;
     }
     
-    get isPublicationEnd(): boolean {
+    get canGoForward(): boolean {
         const lastResource = this.pub.readingOrder.items[this.pub.readingOrder.items.length - 1]?.href;
         return this.viewport.progressions.has(lastResource) && this.viewport.progressions.get(lastResource)?.end === 1;
     }
--- a/navigator/src/epub/EpubNavigator.ts
+++ b/navigator/src/epub/EpubNavigator.ts
@@ -1,5 +1,5 @@
 import { EPUBLayout, Link, Locator, Publication, ReadingProgression } from "@readium/shared";
-import { Configurable, ConfigurablePreferences, ConfigurableSettings, LineLengths, VisualNavigator } from "../";
+import { Configurable, ConfigurablePreferences, ConfigurableSettings, LineLengths, VisualNavigator, VisualNavigatorViewport, ProgressionRange } from "../";
 import { FramePoolManager } from "./frame/FramePoolManager";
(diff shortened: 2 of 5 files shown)


navigator/src/epub/EpubNavigator.ts
```diff
@@ -604,12 +633,30 @@ export class EpubNavigator extends VisualNavigator implements Configurable<Confi
...
-        return [this.currentLocator?.locations.position ?? 0, ...(this.lastLocationInView?.locations.position ? [this.lastLocationInView.locations.position] : [])];
+    get isResourceStart(): boolean {
+        const currentProgression = this.viewport.progressions.get(this.currentLocation?.href);
+        return currentProgression?.start === 0;
+    }
+    
+    get isResourceEnd(): boolean {
+        const currentProgression = this.viewport.progressions.get(this.currentLocation?.href);
+        return currentProgression?.end === 1;
+    }
```
mickael-menu: What are you using `isResourceStart/End` for? Does that still work if several resources are visible on the screen (e.g. fixed layout with spreads or maybe one day continuous scrolling)?

mickael-menu: By the way @qnga used `canGoBackward` and `canGoForward` for `isPublicationStart` / `isPublicationEnd`. Not sure what's best, `isPublicationStart` is more explicit / less opinionated but `canGoForward` fits nicely with `goForward()` and is natural to use to grey out the next/previous buttons.

JayPanoz: `isResourceStart/End` is for the scroll affordances, as we could only rely on positions so far to trigger the top/bottom bars with links to navigate through the publication in Thorium Web – so it could trigger excessively early on smaller devices for instance, since the last position in the resource was far from the bottom of the scrolling element. 

It works for EpubNavigator as it is designed currently, pretty much.  FXL should probably have its own set of helpers though, because the logic is different. Continuous scroll definitely not, we have talked about this in passing multiple times and know it will be complex to get right, with probably many changes so I’m not sure what I could design today would actually cover it in the end. 

> `canGoBackward` and `canGoForward` for `isPublicationStart/isPublicationEnd`

I think I prefer this too, because it leaves room for making something else with `isPublicationStart/End`.

navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
```diff
@@ -39,7 +39,14 @@ export class ColumnSnapper extends Snapper {
     }
 
     reportProgress() {
-        this.comms.send("progress", { progress: this.wnd.scrollX / this.cachedScrollWidth, reference: this.wnd.innerWidth / this.doc().scrollWidth });
+        const scrollX = this.wnd.scrollX;
+        const scrollWidth = this.cachedScrollWidth;
+        const progress = Math.max(0, Math.min(1, scrollX / scrollWidth));
```
chocolatkey, line 44: I'm curious, why did you need to clamp the values to 0-1? When does it go outside of this range?

chocolatkey (approved):

navigator-html-injectables/src/modules/snapper/ColumnSnapper.ts
JayPanoz: @qnga recently discovered an issue with the virtual columns being impacted by rounding errors on Android – this has been corrected since. So I preferred to clamp with that in mind.

navigator/src/epub/EpubNavigator.ts
```diff
@@ -604,12 +633,32 @@ export class EpubNavigator extends VisualNavigator implements Configurable<Confi
...
+    
+    get canGoBackward(): boolean {
+        const firstResource = this.pub.readingOrder.items[0]?.href;
+        return this.viewport.progressions.has(firstResource) && this.viewport.progressions.get(firstResource)?.start === 0;
+    }
+    
+    get canGoForward(): boolean {
+        const lastResource = this.pub.readingOrder.items[this.pub.readingOrder.items.length - 1]?.href;
+        return this.viewport.progressions.has(lastResource) && this.viewport.progressions.get(lastResource)?.end === 1;
```
JayPanoz: @mickael-menu I renamed helpers for Publication and made the resource’s more consistent – they are now using `readingOrder` in order to be compatible with FXL. In theory it should also be compatible with continuous scroll as well – although it might need fine-tuning depending on how it would be implemented.

That said, I am not really happy with their name now, because they are not specifically tied to a single `resource` but the items in `readingOrder` – so that is currently the spread in FXL, the single resource in reflow but could become the loaded iframes in continuous scroll. 

Any idea that could fit across toolkits?

mickael-menu: I agree the name is a bit ambiguous. But maybe we don't need helpers for these in the toolkit directly, at least for now?

JayPanoz: I’ll refer to @HadrienGardeur then, to check whether he prefers to put them in ts-toolkit or Thorium Web.

HadrienGardeur: We've seen that this is a pretty common need since we don't implement continuously scrolling at the moment. I see little harm in having them in the toolkit as simple helpers.

mickael-menu: In that case, would `isScrollStart` / `isScrollEnd` fit better?

JayPanoz: Sounds reasonable to me because I’d like to separate these from the concept of resource if they go in ts-toolkit

JayPanoz:
@mickael-menu I updated the name of the resource/scroll helpers. 

Can this be merged or do you think there are other items to address? 

It is kinda blocking for Thorium Web v1.

mickael-menu (approved):
@JayPanoz It's fine by me. By the way you don't need to wait for me if you need something merged on this repo, we can always adjust APIs before the next release.

JayPanoz:
Thanks! 

Noted. This wasn’t urgent (since #138 is blocking as well and is a lot more complex) so I thought it would be a good idea to double-check beforehand. 🙏

PR #142, opened 2025-07-05
Support for accessibility in shared models

JayPanoz:
This adds support for `accessibility` in `metadata` of Shared Models ([RWPM doc](https://readium.org/webpub-manifest/contexts/default/#accessibility-metadata)). 

Display Guide to follow


JayPanoz:
@chocolatkey I am not anticipating changes myself so I’m marking this as ready to review so that another pair of eyes can see whether this is acceptable or whether we want to do some things in a smarter way. 

As a recap, this adds `accessibility` in `metadata`, which is the most trivial part.
 
It also handles the Accessibility Metadata Display Guide, in a way that should be consistent with Swift – so statements with an `id`, a `compact` string, and a `descriptive` string. 

For localization, I had to design something around thorium-locales. Basically:

- there is a script extracting what we need from the json in order to build its own in `publication/accessibility/locales/`. 
- there is a `Localization` instance that is a singleton, which allows you to register your own locale, setting the one that should be used by classes, and get the compact and descriprive strings.
- At this point it means the json is packaged since we only have `en` and `fr` – as-is Readium CSS in Navigator for instance –, but in theory, Vite should allow us to handle this dynamically as locales/languages are added.

It is kind of straightforward as classes are just mapping statements to conditions but of course there may be room for improvements and/or details that could be polished.

shared/src/publication/accessibility/Accessibility.ts
```diff
@@ -0,0 +1,456 @@
...
+    public static deserialize(json: any): AccessibilityProfile | undefined {
+        if (!json || typeof json !== 'string') return;
+        return new AccessibilityProfile(json);
+    }
+
+    public serialize(): any {
+        return this.uri;
+    }
+
+    public static readonly EPUB_A11Y_10_WCAG_20_A = new AccessibilityProfile('http://www.idpf.org/epub/a11y/accessibility-20170105.html#wcag-a');
```
chocolatkey, line 188: It might be nice to add comments to these for intellisense

shared/src/publication/accessibility/Accessibility.ts
```diff
@@ -0,0 +1,456 @@
...
+    public static deserialize(json: any): AccessMode | undefined {
+        if (!json || typeof json !== 'string') return;
+        return new AccessMode(json);
+    }
+
+    public serialize(): any {
+        return this.value;
+    }
+
+    public static readonly AUDITORY = new AccessMode('auditory');
```
chocolatkey, line 346: Same goes for here (and below), it might be nice to add comments to these for intellisense

shared/src/publication/accessibility/Accessibility.ts
```diff
@@ -0,0 +1,456 @@
...
+        if (!json || typeof json !== 'string') return;
+        return new Feature(json);
+    }
+
+    public serialize(): any {
+        return this.value;
+    }
+
+    public static readonly NONE = new Feature('none');
+    public static readonly ANNOTATIONS = new Feature('annotations');
```
chocolatkey, line 508: @HadrienGardeur should I be adding the missing features to the go-toolkit?

shared/src/publication/accessibility/Accessibility.ts
```diff
@@ -0,0 +1,456 @@
...
+    public static readonly NONE = new Exemption('none');
+    public static readonly DOCUMENTED = new Exemption('documented');
+    public static readonly LEGAL = new Exemption('legal');
+    public static readonly TEMPORARY = new Exemption('temporary');
+    public static readonly TECHNICAL = new Exemption('technical');
+
+    public static readonly EAA_DISPROPORTIONATE_BURDEN = new Exemption('eaa-disproportionate-burden');
+    public static readonly EAA_FUNDAMENTAL_ALTERATION = new Exemption('eaa-fundamental-alteration');
+    public static readonly EAA_MICROENTERPRISE = new Exemption('eaa-microenterprise');
+    public static readonly EAA_TECHNICAL_IMPOSSIBILITY = new Exemption('eaa-technical-impossibility');
```
chocolatkey, line 854: This is also not in the go-toolkit @HadrienGardeur

chocolatkey (approved):
The circular dependency on thorium for the locales data is a bit unusual and couples it strongly to the EDRLab repo, but it works

shared/src/publication/accessibility/Accessibility.ts
JayPanoz: That's a good point will do

PR #144, opened 2025-07-10
Replace presentation with layout in Metadata

JayPanoz:
`Presentation` has been superseded in favour of `Layout`

See RWPM’s [default context](https://readium.org/webpub-manifest/contexts/default/#layout-and-reading-progression) + [EPUB Profile](https://readium.org/webpub-manifest/profiles/epub.html#appendix-b---deprecated-properties)

Still in draft as it depends on https://github.com/readium/go-toolkit/issues/243


JayPanoz:
@chocolatkey I saw in your Go PR that you replaced `auto` with `none` for `ReadingProgression`. I guess I should do this as well then?

chocolatkey:
@JayPanoz Yes, the only options for progression now in the WebPub manifest are `rtl` and `ltr`. "None" in the go-toolkit is just an alias for a blank value, which in the JSON representation means the key would be omitted. Here is the helper to determine the effective progression in Go: https://github.com/readium/go-toolkit/blob/develop/pkg/manifest/metadata.go#L92

navigator/src/epub/fxl/FXLSpreader.ts
```diff
@@ -44,21 +57,21 @@ export class FXLSpreader {
...
+            const orientation = single.properties?.otherProperties["orientation"];
 
             // First page is landscape/spread means no shift
-            if(index === 0 && (orientation === Orientation.landscape || (orientation !== Orientation.portrait && ((single.width || 0) > (single.height || 0) || single.properties?.getSpread() === Spread.both))))
+            if(index === 0 && (orientation === Orientation.landscape || (orientation !== Orientation.portrait && ((single.width || 0) > (single.height || 0) || single.properties?.otherProperties["spread"] === Spread.both))))
                 this.shift = false;
 
             // If last was a true single, and this spread is a center page (that's not special), something's wrong
-            if(wasLastSingle && single.properties?.getPage() === Page.center) {
+            if(wasLastSingle && single.properties?.otherProperties["page"] === Page.center) {
```
chocolatkey: You should still be able to use the page helper here

navigator/src/epub/fxl/FXLSpreader.ts
```diff
@@ -22,8 +35,8 @@ export class FXLSpreader {
                 });
                 // if(!orientation) item.Properties.Orientation = item.Width > item.Height ? "landscape" : "portrait";
             }
-            const isLandscape = item.properties?.getOrientation() === Orientation.landscape ? true : false;
-            if((!item.properties?.getPage() || redo)) item.properties = item.properties?.add({
+            const isLandscape = item.properties?.otherProperties["orientation"] === Orientation.landscape ? true : false;
```
chocolatkey, line 38: Considering the removal of `orientation`, this logic will now no longer work. We can leave this for now, but this code needs to be revamped in the future. Same goes for `spread` below

navigator/src/epub/EpubNavigator.ts
```diff
@@ -117,23 +117,23 @@ export class EpubNavigator extends VisualNavigator implements Configurable<Confi
         this.resizeObserver.observe(this.container.parentElement || document.documentElement);
     }
 
-    public static determineLayout(pub: Publication): EPUBLayout {
-        const presentation = pub.metadata.getPresentation();
-        if(presentation?.layout == EPUBLayout.fixed) return EPUBLayout.fixed;
+    public static determineLayout(pub: Publication): Layout {
```
chocolatkey: This function needs to be updated to take into account the addition of `scrolled` layout, as well as the updated WebPub spec docs. Here is the Go implementation: https://github.com/readium/go-toolkit/blob/develop/pkg/manifest/layout.go#L43

chocolatkey (changes requested):
In addition to my review comments, the answer to your question also should be addressed

navigator/src/epub/fxl/FXLSpreader.ts
JayPanoz: This has been resolved. Given it’s not longer in EPUB (extension) and is now part of the `Properties` proper I renamed it to `page` though, to follow the patterns used in the original codebase.

navigator/src/epub/fxl/FXLSpreader.ts
JayPanoz: Yeah we discussed this with Hadrien a few days ago. The timing for modifying was quite unfortunate though, as I’ll be absent in the next two weeks and stability for v1 of Thorium Web/playground was the best option considering all we had to achieve still.

navigator/src/epub/EpubNavigator.ts
JayPanoz: So I am now relying on this to handle paginated/scroll, instead of `readingProgression`. It probably makes the most sense for future developments of the Navigator.

chocolatkey (approved):
This will work for now

PR #146, opened 2025-07-11
Implement ReadiumCSS iOS patch

JayPanoz:
Depends on https://github.com/readium/css/pull/190

To sum things up, we have a weird unexplained issue on iOS where orientation change bump the size of paragraphs specifically in scroll… 

Since we cannot use `-webkit-text-size-adjust: 100%` because it breaks CSS `zoom` we have to replace the latter with the former on iOS. 

Readium CSS will expose a flag once the PR mentioned above is merged and published. 

This TS-Toolkit PR exposes this flag through the Preferences API, and handles it by default by sniffing iOS + Mobile, as it already does for the iPadOS Patch.

### DIFF
--- a/navigator/src/epub/css/Properties.ts
+++ b/navigator/src/epub/css/Properties.ts
@@ -66,6 +66,7 @@ export interface IUserProperties {
   fontWidth?: FontWidth | null;
   invertFilter?: boolean | number | null;
   invertGaijiFilter?: boolean | number | null;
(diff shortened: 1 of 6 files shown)


JayPanoz:
Not requesting a review cos this is similar to the iPadOSPatch that was added into the Preferences API PR. 

We will rediscuss this when aligning Preferences API across Toolkits so a proper review will be done at this point. 🙏

PR #148, opened 2025-07-15
Scroll fixes

JayPanoz:
This resolves rounding issues on some Android devices re. progression that could never reach `1` in scroll, hence making the `isScrollEnd` helper useless.


JayPanoz:
@chocolatkey Required a review in case you want to handle this differently, especially as there is a discrepancy between scroll and column snappers. 🙏

chocolatkey (approved):

PR #149, opened 2025-07-15
Add missing models in Shared

JayPanoz:
This PR aims to add all the missing models from shared, as accessibility (see #142) is not the only thing missing.


shared/src/publication/divina/Properties.ts
```diff
@@ -0,0 +1,18 @@
...
+// Divina extensions for link [Properties].
+// https://github.com/readium/webpub-manifest/blob/master/schema/extensions/divina/properties.schema.json
+
+declare module '../Properties' {
+  export interface Properties {
+    /**
+     * Specifies that an item in the reading order should break the current continuous scroll 
+     * and start a new one.
+     */
+    getBreakScrollBefore(): boolean;
```
HadrienGardeur: This is being dropped in a PR for Divina, so let's skip this one and properties for Divina in general.

JayPanoz: OK good to know thanks, will remove.

shared/src/publication/GuidedNavigation.ts
```diff
@@ -75,57 +149,91 @@ export class GuidedNavigationObject {
         children?: GuidedNavigationObject[];
         imgref?: string;
         role?: Set<string>;
-        text?: string;
+        level?: number;
+        text?: GuidedNavigationText;
         textref?: string;
+        description?: string;
```
HadrienGardeur: Description can be a Guided Navigation Object without:

- `level`
- and `children`

This means that we can support a textual, auditory or visual description.

HadrienGardeur (changes requested):

shared/src/publication/GuidedNavigation.ts
JayPanoz: Thanks updated.

HadrienGardeur (approved):
LGTM

JayPanoz:
Will merge today unless someone spots an issue then.

PR #152, opened 2025-07-21
Scroll listener

### LINKED ISSUE #151: Expose a scroll listener
Before the Preferences API, the `scroll` event wasn’t even reporting progress as the `ScrollSnapper` did not really care about it. That is why there is no `scroll` event in listeners. 

Now that it is, it would be a good idea to make it a part of `EpubNavigatorListeners` (see also #63) because it will help implement common affordances such as showing/hiding top and bottom bars in scroll, popping up affordances to help streamline navigation through the publication, etc. 

At the moment, `positionChanged` is the only listener that can be used to handle such affordances, but it has some issues that make it more complex than it should actually be. It indeed fires twice on initial load with different progressions so you have to get around that, you cannot really implement a threshold reliably since progression is relative to the current resource, etc. 

I am not sure what `data` the callback can use is at the moment, I’ll have to implement it first.

JayPanoz:
This adds `scroll` in `EpubNavigatorListeners`, which allows to pass a callback with `delta` as an argument. 

Resolves #151

The implementation is less trivial that you would maybe expect, as there is some filtering happening in the `ScrollSnapper`:

- on resize;
- on the initial `scroll` event firing, since we sync the position after mounting the snapper. 

It also removes the previous “scrollEnd” implementation as it was not satisfactory to resolve the root issue we were trying to solve: you had to use `progressionChanged` as a proxy to the scroll event, with side effects and extra complexities, and no way to apply a threshold if needed.

### DIFF
--- a/navigator-html-injectables/src/modules/snapper/ScrollSnapper.ts
+++ b/navigator-html-injectables/src/modules/snapper/ScrollSnapper.ts
@@ -13,13 +13,19 @@ export class ScrollSnapper extends Snapper {
     private wnd!: ReadiumWindow;
     private comms!: Comms;
     private resizeObserver!: ResizeObserver;
+
+    private initialScrollHandled = false;
     private isScrolling = false;
+    private lastScrollTop = 0;
+    private isResizing = false;
+    private resizeDebounce: number | null = null;
 
     private doc() {
         return this.wnd.document.scrollingElement as HTMLElement;
     }
 
     private reportProgress() {
+        if (!this.comms.ready) return;
         // We have to round up the scroll position because
         // Android may never reach 100% of the scroll height
         // due to the way it rounds scrollTop…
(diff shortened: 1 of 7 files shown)


navigator-html-injectables/src/modules/snapper/ScrollSnapper.ts
```diff
@@ -36,29 +41,52 @@ export class ScrollSnapper extends Snapper {
...
+        if (!this.isScrolling) {
+            this.isScrolling = true;
+            this.wnd.requestAnimationFrame(() => {
+                this.reportProgress();
+
+                const currentScrollTop = this.doc().scrollTop;
+                const deltaY = currentScrollTop - this.lastScrollTop;
+                this.lastScrollTop = currentScrollTop;
+
+                this.comms.send("scroll", deltaY);
```
chocolatkey, line 71: How often does this get sent in practice when a user is scrolling? I'm just a little bit worried about not having debouncing here.

chocolatkey (approved):

navigator-html-injectables/src/modules/snapper/ScrollSnapper.ts
JayPanoz: I also thought about throttling this for sure then decided not to as `requestAnimationFrame` did the job pretty well by itself and it wasn’t spamming as much as when you are using a `ResizeObserver` for example – I guess browsers are very cautious when it comes to `scroll` given the performance penalty that can occur. 

But I can add this no problem if you think this would be better.

chocolatkey: We can leave it for now. I'm mainly thinking of future use-cases in e.g. the mobile toolkits where the message passing between native and a webview using the injectables may not be as performant

PR #157, opened 2025-08-29
[Chore] Update deps

JayPanoz:
An update/upgrade of dependencies was long overdue as we were 3 major versions behind Vite latest for instance. 

As far as I can tell everything should test and build fine but it’d be great to have some confirmation. 

It could also be useful to discuss what to make of the vanilla testapp tbh. I did not update the dependencies for this one as I do not know how it will impact it – it is severely out of date at the moment and my focus for the implementation of ts-toolkit has been Thorium Web/Readium Playground so virtually no time has been allocated to it in the last year.

### DIFF
--- a/navigator-html-injectables/package.json
+++ b/navigator-html-injectables/package.json
@@ -34,9 +34,9 @@
   "types": "./types/src/index.d.ts",
   "exports": {
     ".": {
+      "types": "./types/src/index.d.ts",
       "import": "./dist/index.js",
-      "require": "./dist/index.umd.cjs",
-      "types": "./types/src/index.d.ts"
+      "require": "./dist/index.umd.cjs"
     }
   },
   "files": [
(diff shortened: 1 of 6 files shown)


chocolatkey (approved):
It's working on my end as well. Thanks for updating things.

Regarding the vanilla testapp, I'm not really sure what to do with it myself. I feel like it's not a bad idea to have an example of the bare minimum code needed to use the ts-toolkit navigator without having to resort to the more complicated thorium-web, but:
- I don't think either of us have time to continuously maintain it right now
- It's currently based on the now-abandoned material-web
- Realistically, the amount of people trying to build their own reader based on the ts-toolkit instead of based on thorium web is very small

JayPanoz:
Thanks! 

Yeah, this is my exact thoughts, there’s also always the risk of the complicated app having issues a simpler one would perhaps not have or would make the root of this issue more obvious – for instance, it’s been very difficult to debug the Webkit-specific issues re. scroll and narrow these down to a minimal reproducible test case. 

But there is no easy way around the things you listed. Although a test app, this is still an app that requires development time.

PR #159, opened 2025-09-09
fix: replace instanceof checks with more appropriate ones

tefkah:
This PR replaces ~10 `instanceof Array` checks with `Array.isArray`, and two `instanceof Object` checks with `typeof x === 'object'`)

This is because I ran into the following issue when trying to test the `decorations`.

Say i have a button like so (assuming i have access to the `navigator` in some way)

```tsx
 <Button
              onClick={async () => {

                const positionLocator = new Locator({
                  href: "some href",
                  type: "application/xhtml+xml",
                  locations: new LocatorLocations({
					// just an example
                    fragments: [`c09-sentence0`],
                  }),
                })

                navigator.send("decorate", {
                  group: "highlight",
                  action: "add",
                  decoration: {
                    id: "highlight",
                    locator: positionLocator,
                    style: {
                      tint: "red",
                      width: Width.Wrap,
                      layout: Layout.Bounds,
                    },
                  },
                })
           }}
>
              Manual higlhight
 </Button>
```

When `arrayFromJsonOrString` is called here on `positionLocator.locations.fragments`

https://github.com/readium/ts-toolkit/blob/94e518c3a43e624d82063ac4e864149aba787aad/shared/src/publication/Locator.ts#L70

it will always return `undefined`, because the `Array` referenced in `instanceof Array` is `navigator._cframes[number].window.Array`, not window `Array`

You can see this by doing the following in the console a page with an `iframe` , like here: https://playground.readium.org/read/moby-dick

```js
arr = ['a', 'b']; arr instanceof (document.querySelector('iframe').contentWindow).Array  // false
```

vs doing

```js
arr = ['a', 'b']; (document.querySelector('iframe').contentWindow).Array.isArray(arr) // true
```

The same goes for `instanceof Object`.

On `instanceof Object`, you might worry that `typeof x === 'object'` does not cover the same ground. In the cases where I replaced this, I'm quite confident it does however.

## Caveats

It might be possible that this is somehow intended, but I don't think so, given the subtle nature of the bug (took me a couple hours to figure out).

### DIFF
--- a/shared/src/publication/BelongsTo.ts
+++ b/shared/src/publication/BelongsTo.ts
@@ -25,7 +25,7 @@ export class BelongsTo {
    * Parses a [BelongsTo] from its RWPM JSON representation.
    */
   public static deserialize(json: any): BelongsTo | undefined {
-    if (!(json && json instanceof Object)) return;
+    if (!(json && typeof json === 'object')) return;
 
     const items = new Map<string, Collection>();
     Object.entries(json).forEach(([key, value]) => {
--- a/shared/src/publication/PublicationCollection.ts
+++ b/shared/src/publication/PublicationCollection.ts
@@ -42,7 +42,7 @@ export class PublicationCollection {
     if (Array.isArray(json)) {
       // Parses an array of links.
       links = Links.deserialize(json);
-    } else if (json instanceof Object) {
+    } else if (typeof json === 'object') {
       // Parses a sub-collection object.
       links = Links.deserialize(json.links);
 
--- a/shared/src/publication/PublicationCollection.ts
+++ b/shared/src/publication/PublicationCollection.ts
@@ -39,7 +39,7 @@ export class PublicationCollection {
     let metadata: Map<string, any> | undefined;
     let subcollections: Map<string, Array<PublicationCollection>> | undefined;
 
-    if (json instanceof Array) {
+    if (Array.isArray(json)) {
       // Parses an array of links.
       links = Links.deserialize(json);
     } else if (json instanceof Object) {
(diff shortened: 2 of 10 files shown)


shared/src/publication/BelongsTo.ts
```diff
@@ -25,7 +25,7 @@ export class BelongsTo {
    * Parses a [BelongsTo] from its RWPM JSON representation.
    */
   public static deserialize(json: any): BelongsTo | undefined {
-    if (!(json && json instanceof Object)) return;
+    if (!(json && typeof json === 'object')) return;
```
tefkah, line 28: this check is identical, except it works across `iframe` boundaries.

shared/src/publication/PublicationCollection.ts
```diff
@@ -39,10 +39,10 @@ export class PublicationCollection {
     let metadata: Map<string, any> | undefined;
     let subcollections: Map<string, Array<PublicationCollection>> | undefined;
 
-    if (json instanceof Array) {
+    if (Array.isArray(json)) {
       // Parses an array of links.
       links = Links.deserialize(json);
-    } else if (json instanceof Object) {
+    } else if (typeof json === 'object') {
```
tefkah, line 45: same here: since we have already checked that `!!json` and `!Array.isArray(json)`, we can be sure that we have a regular ol object on our hands

shared/src/publication/BelongsTo.ts
tefkah: it doesn't distinguish between `object`s and `Array`s, but the original check also didn't

JayPanoz:
Thanks for catching this, especially as we've been updating shared models lately and that slipped through the cracks. 

I will review it as soon as possible but at first sight I don't see anything problematic, and I am not aware of any particular reason why that was used in the first place – I was indirectly involved in the creation of shared models and can't recall nor can find any reason in writing why that was used instead of your changes at least.

JayPanoz (approved):
LGTM. 

Can confirm tests are passing as expected and this should not have any unintended side-effects as we’re always checking json is truthy/falsy before the other checks. 

Will merge with the other PRs removing the post install script and the ReadiumCSS update and publish a `2.1.1` for all packages since the shared modifications are impacting all other components of the monorepo. 

Thanks again for catching this, this is greatly appreciated. 👍

PR #160, opened 2025-09-09
fix: remove postinstall hook for @readium/shared

tefkah:
When running

```sh
npm i @readium/shared@2.1.0
```

the install fails due to a missing file necessary for the `postinstall` hook, see full error below.

This is because `update-accessibility-locales.mjs` is not included in the `npm` bundle.

Although this issue could also be solved by including it, it doesn't appear that the resulting files are necessary for the consumer, only for the developer. Therefore I suggest removing it, and finding another way (a commit/push hook eg) to make the `update-accessibility-locales` script work.

This may have gone undetected bc `pnpm` does not run `postinstall` scripts by default, but `yarn` and `npm` do.

```log
npm error code 1
npm error path ~/Projects/test-readium-shared-install/node_modules/@readium/shared
npm error command failed
npm error command sh -c npm run update-accessibility-locales
npm error > @readium/shared@2.1.0 update-accessibility-locales
npm error > node update-accessibility-locales.mjs
npm error npm warn Unknown env config "python". This will stop working in the next major version of npm.
npm error npm warn Unknown user config "python". This will stop working in the next major version of npm.
npm error node:internal/modules/cjs/loader:1372
npm error   throw err;
npm error   ^
npm error
npm error Error: Cannot find module '~/Projects/test-readium-shared-install/node_modules/@readium/shared/update-accessibility-locales.mjs'
npm error     at Module._resolveFilename (node:internal/modules/cjs/loader:1369:15)
npm error     at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
npm error     at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
npm error     at Module._load (node:internal/modules/cjs/loader:1179:37)
npm error     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
npm error     at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
npm error     at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:152:5)
npm error     at node:internal/main/run_main_module:33:47 {
npm error   code: 'MODULE_NOT_FOUND',
npm error   requireStack: []
npm error }
npm error
npm error Node.js v24.4.1
npm error A complete log of this run can be found in: ~/.npm/_logs/2025-09-09T13_37_12_939Z-debug-0.log
```

### DIFF
--- a/shared/package.json
+++ b/shared/package.json
@@ -48,7 +48,6 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "update-accessibility-locales": "node update-accessibility-locales.mjs",
-    "postinstall": "npm run update-accessibility-locales",
     "test": "jest",
     "size": "size-limit"
   },


JayPanoz:
Apologies, that was my oversight, it should have been `prepare` – which does not run when installing as a dependency –, although Yarn does not support it if I am not mistaken, but we are committing the files anyway so it is not that critical. 

For context, these are generated for dev purposes as you correctly guessed, for the Accessibility Metadata Display Guide localization. To sum things up this class takes accessibility metadata and converts it into legible statements (strings) end-users can understand more easily, using common terms. 

That is also possibly temporary as we discuss whether we should convert the Thorium locales repo into a package proper. So I'm not sure we want to replace it with a prepare script anyway. Any feedback would be welcome as it is kind of an unusual dependency

tefkah:
ah i see!

i have successfully used [`gitpkg`](https://gitpkg.vercel.app/about) for that in the past, altho it's not a solution i'd rely on for too long (no way of knowing how long it will stay available!)

but it's quite nice, you can just point it to a certain subdirectory of a git dir and turn it into an installable npm package.

i gave it a shot and it seems to work!

if i add

```json
"dependencies": {
    "@thorium/locales": "https://gitpkg.vercel.app/edrlab/thorium-web/public/locales?develop"
},
```

to `package.json`

and 
```ts
  resolve: {
    alias: {
      "@thorium/locales": resolve(
        __dirname,
        "../node_modules/.pnpm/locales+develop@https+++gitpkg.vercel.app+edrlab+thorium-web+public+locales+develop/node_modules/locales?develop"
      ),
    },
  },
```

to `vite.config.js` (sadly necessary as `pnpm` kind of messes up the install path)

then i'm able to replace

https://github.com/readium/ts-toolkit/blob/766c6b60fbc919db7424a268f56d33d1a1b18adc/shared/src/publication/accessibility/Localization.ts#L1-L3

with 

```ts
// Localization.ts
import enUS from '@thorium/locales/en/thorium-web.json'
import frFR from '@thorium/locales/fr/thorium-web.json'
```

and sucessfully build.


If you want I can modify my PR to do this instead!

I would suggest spinning off the `locales` into a separate package in the future, but this can work until that time.

JayPanoz:
Thanks for the insight that is interesting. 

Your offer is highly appreciated but I'd say we probably shouldn't worry too much about that since we are committing the generated files anyway. At least that makes the life of contributors and users simpler by having something stable that won't break unexpectedly on auto-update. 

Ideally we should turn the Thorium locales into a package, as you stated. I prefer you do not spend additional time on this but on your own projects. : 🙏 

Manual updates are very manageable at this point in time and we know that the localization system will change quite a bit as locales are added to the repository so that we do not have to worry about imports, the package size, etc. So we'll come back to it eventually.

Issue #120, opened 2025-05-24
Hardening EPUB content in iframes


chocolatkey:
EPUB content, as is valid per the latest EPUB spec, is inherently dangerous as creators of EPUB files have the ability to take full advantage of the web browser's HTML, CSS and Javascript engines. With the way the ts-toolkit is currently implemented, it is an easy task for EPUB content in an iframe create and managed by the toolkit to escape and execute malicious JS on the parent frame. Any data in localstorage, or cookies that are not fully secured, are fair game, as well as other behavior such as deceptive UIs, tracking of user data, location etc.
While most large users of the ts-toolkit will be receiving their publications from trusted sources, this is not always the case, and we should not expect users of the toolkit to be aware of all the risks that come with accepting user-generated content in their web reader. Therefore, we should take the necessary steps to lock down content as much as possible by default, possibly providing ways to disable these protections if explicitly specified.
The EPUB spec lists some of the concerns here: https://www.w3.org/TR/epub-33/#sec-security-privacy
Here is an EPUB the author of [foliate](https://github.com/johnfactotum/foliate) made to test many of the security issues: https://github.com/johnfactotum/epub-test/tree/master
I think we should take a stricter approach by default than what is allowed in the EPUB spec, as simply loading an external image gives the host of that image the ability to track your approximate location (IP address), web browser, and more.
Personally, I would suggest we completely disallow the loading of remote resources by default. And even if this is explicitly enabled by the implementer, it should be contingent on some sort of property in the webpub manifest, which I would suggest [porting over from the original EPUB](https://www.w3.org/TR/epub-33/#sec-remote-resources).

I have done some experimentation with the ts-toolkit to see what we can do to secure things. Here is what I've come up with so far we could do (note this still restricts some stuff that is allowed in the EPUB spec):

## Injecting CSP into the iframe
During the building of the iframe content, we already inject tags into the `<head>` of the document. The [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy) can be injected not only using HTTP headers (which is very difficult for us to do), but also by using a `<meta>` tag in the head of the document. I have tried this out, and it seems to be pretty effective. Example code snippet:
```ts
// FrameBlobBuilder.ts
// root = the base path of the publication

const csp = (domains: string[]) => {
    const d = domains.join(" ");
    return [
        // 'self' is useless because the document is loaded from a blob: URL
        `upgrade-insecure-requests`,
[...]

PR #169, opened 2025-11-07
iframe hardening

chocolatkey:
Related to #120

### DIFF
--- a/navigator/src/epub/frame/FrameBlobBuilder.ts
+++ b/navigator/src/epub/frame/FrameBlobBuilder.ts
@@ -50,7 +50,10 @@ const cssSelectorGenerator = (doc: Document) => scriptify(doc, cached("css-selec
 
 // Note: we aren't blocking some of the events right now to try and be as nonintrusive as possible.
 // For a more comprehensive implementation, see https://github.com/hackademix/noscript/blob/3a83c0e4a506f175e38b0342dad50cdca3eae836/src/content/syncFetchPolicy.js#L142
+// The snippet of code at the beginning of this source is an attempt at defence against JS using persistent storage
 const rBefore = (doc: Document) => scriptify(doc, cached("JS-Before", () => blobify(stripJS(`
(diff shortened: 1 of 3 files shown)


JayPanoz:
@chocolatkey notwithstanding the scripting issues we are already aware of with demo files, do you think we will end up with something radically different in Web Publications, or is there at least some overlap you can anticipate?

navigator/src/epub/frame/FrameBlobBuilder.ts
```diff
@@ -50,7 +50,10 @@ const cssSelectorGenerator = (doc: Document) => scriptify(doc, cached("css-selec
...
+// The snippet of code at the beginning of this source is an attempt at defence against JS using persistent storage
 const rBefore = (doc: Document) => scriptify(doc, cached("JS-Before", () => blobify(stripJS(`
+    const noop=()=>{},emptyObj={},emptyPromise=()=>Promise.resolve(void 0),fakeStorage={getItem:noop,setItem:noop,removeItem:noop,clear:noop,key:noop,length:0};["localStorage","sessionStorage"].forEach((e=>Object.defineProperty(window,e,{get:()=>fakeStorage,configurable:!0}))),Object.defineProperty(document,"cookie",{get:()=>"",set:noop,configurable:!0}),Object.defineProperty(window,"indexedDB",{get:()=>{},configurable:!0}),Object.defineProperty(window,"caches",{get:()=>emptyObj,configurable:!0}),Object.defineProperty(navigator,"storage",{get:()=>({persist:emptyPromise,persisted:emptyPromise,estimate:()=>Promise.resolve({quota:0,usage:0})}),configurable:!0}),Object.defineProperty(navigator,"serviceWorker",{get:()=>({register:emptyPromise,getRegistration:emptyPromise,ready:emptyPromise()}),configurable:!0});
```
danielweck, line 55: The vast majority of EPUB publications don't need localStorage, so in the real world it probably won't hurt to disable it, but just FYI: disabling `localStorage` kills interactive features in some EPUBs (reflow and fixed layout), for example where the user can enter their name for a personalised story, or where some progress data is stored to deliver a contextual user experience. In Thorium Desktop each publication has its own origin so the store is shared amongst all book chapters but doesn't leak out, there is some built-in isolation already (I guess the web reader cannot make this assumption).

JayPanoz: Yeah I was quite involved in this topic – you may remember I was the one who accidentally discovered Apple Books.app used the same domain for all books and local storage was consequently global/shared, and they had to rush a new release as it went public on Twitter. 😶 

At some point there were talks and some spec draft for suborigins or something like that. 

I'm not holding my breath as it was a long long time ago and I'm not finding anything relevant. 😬

Edit: And well it wouldn't be a toolkit concern anyway.

PR #176, opened 2025-11-24
Readium css experiments

JayPanoz:
This updates ReadiumCSS and adds support for experimental flags/experiments. 

Since ReadiumCSS exposes these as JSON, we are using the key to type and ensure values exist before assigning them. 

This also creates RSProperties in WebPubCSS, as these flags are handled in the `--RS__` scope, not the `--USER__` one, which should explain why it is not found in preferences, but only defaults – and settings.

### DIFF
--- a/navigator/package.json
+++ b/navigator/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "@laynezh/vite-plugin-lib-assets": "^2.1.0",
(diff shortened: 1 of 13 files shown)


JayPanoz:
No review as we’ve done without it for the Preferences API + Readium CSS updates, but when consolidating Navigators once the Web one is no longer experimental, we will make sure to make a very thorough one. 🙏

PR #178, opened 2025-12-16
fix: Metadata.serialize() sortAs and undefined language tag

SifAa:
I discovered that when calling Manifest.serialize() it did not return correctly for sortAs in metadata and that undefined language did not use a correct BCP-47 tag. This PR should fix it.

### DIFF
--- a/shared/src/publication/LocalizedString.ts
+++ b/shared/src/publication/LocalizedString.ts
@@ -9,7 +9,7 @@
 export class LocalizedString {
   public readonly translations: { [key: string]: string };
 
-  public static readonly UNDEFINED_LANGUAGE = 'undefined';
+  public static readonly UNDEFINED_LANGUAGE = 'und';
   public static readonly LANGUAGE_EN = 'en';
 
   /**
--- a/shared/src/publication/Metadata.ts
+++ b/shared/src/publication/Metadata.ts
@@ -287,7 +287,7 @@ export class Metadata {
     if (this.identifier !== undefined) json.identifier = this.identifier;
     if (this.altIdentifier) json.altIdentifier = this.altIdentifier.serialize();
     if (this.subtitle) json.subtitle = this.subtitle.serialize();
-    if (this.sortAs) json.sortAs = this.sortAs.serialize();
+    if (this.sortAs) json.sortAs = this.sortAs.getTranslation();
     if (this.editors) json.editor = this.editors.serialize();
     if (this.artists) json.artist = this.artists.serialize();
     if (this.authors) json.author = this.authors.serialize();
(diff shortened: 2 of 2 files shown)


JayPanoz:
So funny story, I will have to check if there is any side-effect re. `undefined` →  `und` cos’ of the following: 

```js
const test = { undefined: "unknown", en: "english", fr: "french" }
test["en"] // Returns "english"
test[undefined] // Note the lack of quotes. Returns "unknown"
```

So I’ll have to check pretty much all packages for this, to see whether things accidentally worked because `undefined` was used instead of `und` when we didn’t have a defined value.

shared/src/publication/Metadata.ts
```diff
@@ -287,7 +287,7 @@ export class Metadata {
     if (this.identifier !== undefined) json.identifier = this.identifier;
     if (this.altIdentifier) json.altIdentifier = this.altIdentifier.serialize();
     if (this.subtitle) json.subtitle = this.subtitle.serialize();
-    if (this.sortAs) json.sortAs = this.sortAs.serialize();
+    if (this.sortAs) json.sortAs = this.sortAs.getTranslation();
```
JayPanoz: This allowed us to spot an inconsistency in the RWPM schema, where `sortAs` in `Metadata` was effectively correct, but `sortAs` in `Contributor` and `Subject` were not. 

In any case, its `serialization` has to be consistent with `title` and `subtitle` as they are using the same type, `LocalizedString`. 

Note I’ve merged a fix in #183, and updated the schema in https://github.com/readium/webpub-manifest/pull/128#issuecomment-3670229974

Maintainers of the other toolkits have been informed of this change.

JayPanoz:
```suggestion
    if (this.sortAs) json.sortAs = this.sortAs.serialize();
```

shared/src/publication/LocalizedString.ts
```diff
@@ -9,7 +9,7 @@
 export class LocalizedString {
   public readonly translations: { [key: string]: string };
 
-  public static readonly UNDEFINED_LANGUAGE = 'undefined';
+  public static readonly UNDEFINED_LANGUAGE = 'und';
```
JayPanoz, line 12: Note I will have to update tests directly after this is merged, as they currently fail due to `undefined` being used for `LocalizedString`

JayPanoz (changes requested):
Thanks. This helped us spot an inconsistency in RWPM schemas.

So it is expected `sortAs` should be treated as a `LocalizedString`, like `title` and `subtitle`, and that they should then all use the same method for serialisation – otherwise data is lost. 

As for `und` thanks for switching to this. I will update tests following the merge so that you don’t have to as I already did it when reviewing this PR.

shared/src/publication/LocalizedString.ts
JayPanoz: @SifAa Just to double-check I’ve been assuming “Allow edits and secrets from maintainers” is unchecked for this PR as I could not push changes in test to the PR. Hence why I wanted to do it after merging. 

Is that indeed the case or am I missing something in the process? Ideally it would be neater if we could do it directly in the PR but if that is not possible then I’ll follow up with another one.

JayPanoz (approved):
LGTM. Thanks for catching and updating this. 

I’ll take care of tests

PR #179, opened 2025-12-16
feat: inject custom css and js

SifAa:
In the swift and kotlin toolkits we can inject custom CSS and JS. This is to add that feature in the ts-toolkit.

### DIFF
--- a/navigator/src/epub/EpubNavigator.ts
+++ b/navigator/src/epub/EpubNavigator.ts
@@ -9,7 +9,7 @@ import { FXLFrameManager } from "./fxl/FXLFrameManager";
 import { FrameManager } from "./frame/FrameManager";
 import { IEpubPreferences, EpubPreferences } from "./preferences/EpubPreferences";
 import { IEpubDefaults, EpubDefaults } from "./preferences/EpubDefaults";
-import { EpubSettings } from "./preferences";
+import { EpubInjections, EpubSettings } from "./preferences";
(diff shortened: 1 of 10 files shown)


JayPanoz:
Thanks! Very much appreciated. 

For the record we’ve been thinking of and working on an API for injectables – to be implemented cross-platform –, but it’s already a good thing you offer a PR bringing ts-toolkit on par with the mobile toolkits. This will already allow for injecting CSS and JS while prototyping and finalising this API. 

I will make sure to review this PR ASAP, then stay in touch for the subsequent API we will discuss in the upcoming weeks.

JayPanoz:
Alright, thanks for your patience. 

We have decided to not merge this PR as we’ve been working on an Injection API, as mentioned previously. This Injection API is my top priority right now and should be PR’d and merge very soon. It will also address some shortcomings we could notice when reviewing this PR e.g.:

- listing resources by stylesheet/script – versus all resources by default
- prepending or appending – versus appending, as it helps cover more use cases, including ReadiumCSS, but also running scripts before publishers’/authors’
- support for `preload`
- allowing external CSS/scripts in addition to strings
- blob management – if a stylesheet/script is used by multiple resources, why re-create it for each resource, etc. 

Note this API should be implemented across all toolkits, so everything should be consistent across our ecosystem in the end. 

Thanks again for the PR. Even if not merged, it was useful in the sense it could help address core features of the API we are designing. 🙏

PR #184, opened 2026-01-15
feat: add highlight ID attribute to decoration items

oscarhn27:
In the previous version of Readium (ReadiumJs) there was a highlights plugin in which, in addition to adding, editing, and deleting, it was possible to register events when a highlight was clicked or hovered.

Currently, this can be implemented manually by retrieving the boundingRects using the readium-highlight class and manually capturing events in the iframe (although it would be great to implement all of this in the library).

However, what cannot be done is detecting which highlight the selected HTML element refers to, because these elements are empty. Therefore, I propose **adding to the highlight element’s dataset the identifier that the user passes to the method, as an attribute of the highlight.**

### DIFF
--- a/navigator-html-injectables/src/modules/Decorator.ts
+++ b/navigator-html-injectables/src/modules/Decorator.ts
@@ -209,6 +209,7 @@ class DecorationGroup {
 
         const itemContainer = this.wnd.document.createElement("div");
         itemContainer.setAttribute("id", item.id);
+        itemContainer.dataset.highlightId = item.decoration.id;
         // itemContainer.dataset.style = item.decoration.style; // TODO style
         itemContainer.style.setProperty("pointer-events", "none");
 


JayPanoz:
Thanks! 

Will take a look ASAP. 

Discussion #172 may be of interest as well. It is planned I work on the Decorator API at some point and improve it – since we need it for read along. 

So any input is greatly appreciated so that we can implement something for such use cases. 🙏

oscarhn27:
Thank you very much for the quick response, I really appreciate it. I wasn’t aware that there was an ongoing discussion, and I’d be happy to take part in it when I can.

This PR, however, is not so much a comprehensive solution as it is a small patch so that developers can start using this kind of interaction with highlights as soon as possible, in a somewhat “artisanal” way.

I’ll keep an eye on upcoming updates and I’m also happy to help in any way I can. Thanks!

JayPanoz:
Oh sure no worries. It is expected this is not trivial work that will come all at once in a PR. 

Which is why it can be useful to gather interested parties in this discussion so that we can nail an API/helpers down ASAP, given the growing interest in highlighting/decorators. 

I do not see anything problematic given the scope of the change, but I'll just test the PR from the package consumers' point of view as I always do to make sure it is fine.

PR #185, opened 2026-01-19
Chore deps update

### LINKED ISSUE #117: Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Document': '[epub:type]' is not a valid selector.
This can be reproduced while clicking on subtitles in the chapter headings of this [publication for instance](https://playground.readium.org/read?book=https%3A%2F%2Fpublication-server.readium.org%2FbmF0aGFuaWVsLWhhd3Rob3JuZV90aGUtaG91c2Utb2YtdGhlLXNldmVuLWdhYmxlc19hZHZhbmNlZC5lcHVi) e.g.

```
<p epub:type="title">A Day Behind the Counter</p>
```

NextJS gives more details: 

```
SyntaxError: Failed to execute 'querySelectorAll' on 'Document': '[epub:type]' is not a valid selector.
src/readium/ts-toolkit/navigator/dist/index.js (4012:59) @ getCssSelector

  4010 |       targetElement: t.target.outerHTML,
  4011 |       interactiveElement: (r = cA(t.target)) == null ? void 0 : r.outerHTML,
> 4012 |       cssSelector: this.wnd._readium_cssSelectorGenerator.getCssSelector(t.target)
       |                                                           ^
  4013 |     }), this.pointerMoved = !1;
  4014 |   }
```

JayPanoz:
This PR removes the vanilla test app and updates dependencies across packages. 

This closes #117 through https://github.com/readium/ts-toolkit/pull/185/commits/f10ef9330eed61394ab7075d74d02a098492fc69 by scripting the modifications needed in CSS Selector Generator dependencies (Navigator > FrameBlobBuilder)

It adds all colon attributes e.g. `epub:type` in `blacklist` by default when outputting the custom script but maybe we want to scope this. I am not sure at the moment, despite testing the change – because it is tough to encounter edge cases from a limited amount of publications in Thorium Web.


chocolatkey (approved):
- RIP testapp
- Eventually we should come up with a better solution for the css selector code, but should suffice for now

JayPanoz:
Yeah re. css selector, I came across this instance in services: https://github.com/readium/ts-toolkit/blob/118269d10df37a7a979c4e96f749b9d87d51c6e9/shared/src/publication/services/content/iterators/HTMLResourceContentIterator.ts#L327-L360

At first sight it would be tempting to align it on `blacklist` but admittedly testing the change is quite an issue. There has to be a reason it was done like this, and I’m uncomfortable updating it in this PR without the entire context – that would be the FXL Navigator PR so it’s massive in scope, and that type of details might go forgotten.

JayPanoz:
Well actually turns out I do not even need to `blacklist` anything as it’s been fixed in CSS Selector Generator. I was confused by the version being customised and injected, which version was prior to the one in dependencies.

PR #191, opened 2026-01-28
Update Localization module

JayPanoz:
This switches to dynamic imports as we now have more languages in  `publication-metadata` module, with a focus on `accessibility.display-guide`.

This is not using `import.meta.glob` as it is problematic for testing, but what’s implemented results in the same output on `build` so it should not be a significant issue. 

List of supported languages has to be kept updated manually, as we have zero guarantee the available JSON are complete.


JayPanoz:
Eventually we might want to switch to the official `i18next` deps, so that we do not need to care about features and their extended support but this will require additional research and work.

PR #198, opened 2026-03-05
WIP: FXL/Divina improvements

chocolatkey:


navigator/src/epub/frame/FrameBlobBuilder.ts
```diff
@@ -20,73 +20,121 @@ const csp = (domains: string[]) => {
...
+            doc.head.appendChild(viewportMeta);
+        }
+
         const simg = document.createElement("img");
-        simg.src = this.burl || "";
-        simg.alt = this.item.title || "";
-        simg.decoding = "async";
+        simg.src = burl || "";
+        simg.alt = link.title || "";
+        await simg.decode(); // Reduce repaints
```
JayPanoz, line 116: What happens in case of an `EncodingError`? There is no guard here so I’m wondering if it could turn into a hard failure for the whole page build or if that is fine.

navigator/src/epub/EpubNavigator.ts
```diff
@@ -779,6 +798,8 @@ export class EpubNavigator extends VisualNavigator implements Configurable<Confi
     // End of Decoration
 
     public async destroy() {
+        // Flag synchronously so an in-flight load() bails before attaching frames.
+        this._destroyed = true;
```
JayPanoz, line 802: Does WebPubNavigator need this too?

navigator/src/audio/AudioNavigator.ts
```diff
@@ -94,6 +114,8 @@ export class AudioNavigator extends MediaNavigator implements Configurable<Audio
     /** True while a track transition is in progress; suppresses spurious mid-navigation events. */
     private _isNavigating: boolean = false;
     private _isStalled: boolean = false;
+    /** Set by destroy(); stops the deferred initial load from touching a dead instance. */
+    private _destroyed: boolean = false;
```
JayPanoz, line 118: Should this guard mediaElementSetup in addition to startInitialLoad? I am thinking about an edge case of a synchronous `destroy()` running setup against an orphaned element.

navigator/src/divina/preferences/DivinaDefaults.ts
```diff
@@ -0,0 +1,37 @@
...
+import { DivinaQuality } from "../DivinaVariantSelector.ts";
+import { stripWidthRangeConfig } from "./DivinaPreferences.ts";
+
+export interface IDivinaDefaults {
+  backgroundColor?: string | null,
+  constraint?: number | null,
+  quality?: DivinaQuality | null,
+  scrolled?: boolean | null,
+  spreads?: boolean | null,
+  stripWidth?: number | null
```
JayPanoz, line 18: I guess that eventually we want to consolidate this with an additional `fit` preference/setting e.g. fit/cover/contain, etc. – I don’t remember the values exactly. Note a fit preference also applies to FXL so it is not specifically Divina.

navigator/src/divina/preferences/DivinaPreferencesEditor.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  constructor(initialPreferences: DivinaPreferences, settings: DivinaSettings, metadata: Metadata) {
+    this.preferences = initialPreferences;
+    this.settings = settings;
+    this.metadata = metadata;
+  }
+
+  clear(): void {
+    this.preferences = new DivinaPreferences({});
+  }
```
JayPanoz, line 22: Note I recently corrected that, has all values have to be nullified – otherwise they are undefined on merge and it is a no-op

navigator/src/divina/DivinaNavigator.ts
```diff
@@ -0,0 +1,662 @@
...
+import { DivinaPreferencesEditor } from "./preferences/DivinaPreferencesEditor.ts";
+
+export interface DivinaNavigatorConfiguration {
+    preferences: IDivinaPreferences;
+    defaults: IDivinaDefaults;
+    contentProtection?: IContentProtectionConfig;
+    keyboardPeripherals?: IKeyboardPeripheralsConfig;
+}
+
+export interface DivinaNavigatorListeners {
```
JayPanoz, line 40: As a reminder, we need to add `timelineItemChanged` to that list – and in publication as well.

navigator/src/divina/DivinaVariantSelector.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * User-selectable image quality for publications that provide alternate
+ * resolutions of their pages (RWPM `alternate` links).
+ */
+export enum DivinaQuality {
+    auto = "auto", // Match the display resolution
+    low = "low", // Smallest available variant
+    high = "high", // One step above the display resolution
+    max = "max", // Largest available variant
+}
```
JayPanoz, line 13: I guess if feasible it could be useful to have an accompanying helper in shared deriving what’s available from manifest so that consumers can pick the values they provide, chose not to display the setting if there is no variant, etc. 

Of course that means having opinionated heuristics for corner cases.

navigator/src/divina/DivinaNavigator.ts
```diff
@@ -0,0 +1,662 @@
...
+        probe.style.display = "none";
+        probe.style.color = color;
+        doc.body.appendChild(probe);
+        const rgb = (doc.defaultView || window).getComputedStyle(probe).color;
+        probe.remove();
+        const m = rgb.match(/rgba?\(\s*(\d+)[,\s]+(\d+)[,\s]+(\d+)/);
+        if(!m) return null;
+        const luminance = (0.2126 * +m[1] + 0.7152 * +m[2] + 0.0722 * +m[3]) / 255;
+        return luminance < 0.45;
+    }
```
JayPanoz, line 404: I guess that if it does not differ too much – that could use the color helpers from the helpers package as we are checking dark/light in Decorator module already.

Of course the color helper itself could be improved and/or updated for versatility.

JayPanoz:
Some first inputs, more than a proper review as there is some catching up to do and the PR is still in draft. Overall, I do not see anything problematic. On the opposite, it resolves some issues. 

Preferences API is implemented as expected, and the `fit` preference I am mentioning is something we wanted to implement for FXL already so I guess it could be handled at this point. Strictly, if we display images directly e.g. light novel, I guess it wouldn’t necessarily be useless to provide it as a preference in reflowable either. 

The biggest visible gap is the absence of Timeline API but this is to be expected as well, as it is fairly new in the develop branch. Other than that, looks very positive to me, especially after testing the implementation in a reading app.

navigator/src/epub/frame/FrameBlobBuilder.ts
chocolatkey: That is a good question, I haven't accounted for that. I think the error would propagated up into the FramePoolManager where there's a try/catch for `.build`, and further from there. There are other exceptions can be thrown in the FrameBlobBuilder functions, so basically it would fail like they do. Would you like a different behavior? I think a decode failure is fatal

navigator/src/epub/EpubNavigator.ts
chocolatkey: I did not touch the WebPubNavigator at the time because it was in active development by you and I didn't want to meddle with it. So it's up to you

navigator/src/audio/AudioNavigator.ts
chocolatkey: @JayPanoz please see if my latest commit does what you are suggesting

navigator/src/epub/frame/FrameBlobBuilder.ts
JayPanoz: I’m personally fine as long as it is consistent, but we have issue #257 demonstrating that maybe there will be subsequent work around error handling in FrameBlobBuilder.

navigator/src/divina/preferences/DivinaDefaults.ts
chocolatkey: I agree. This is modeled mainly on what uses are used to on webtoon-reading sites as a setting

navigator/src/divina/preferences/DivinaPreferencesEditor.ts
chocolatkey: Do you want to address this in your conflict resolution or should I?

navigator/src/epub/EpubNavigator.ts
JayPanoz: WebPubNavigator should be fine for the time being and I won’t touch it further. If anything I’d like the next step consolidating it with EPUB reflow as there is lots of duplicated code.

navigator/src/divina/preferences/DivinaPreferencesEditor.ts
JayPanoz: I’ll do. No worries. 

I shall encounter it when I walk up the PRs/diff.

navigator/src/epub/frame/FrameBlobBuilder.ts
chocolatkey: Personally what I want to do after having had users reporting issues on my platform, is to use the SVG generator for the divina pages which currently just shows a generic error page when the page fails to load, is to come up with an easy to see error code system for these different errors, so they can be displayed in a large enough font, screenshotted by end-users, and forwarded to devs. So instead of "Error" or a symbol, there would be something like "Error: CL_0123", where the prefix is the major component and the number is a sequential one assigned as we wrap errors with this. I was going to do this post-merge. Of course the stack would still be in the console for us

JayPanoz: That makes sense. Thanks for the update.

JayPanoz:
Alright turns out conflict resolution was not as complex as months of parallel development would lead you to believe. I must still test to make sure everything is fine – as you probably do – but there wasn’t like a huge and hairy chunk of code to reconcile. 

Note I published a backup just in case here: https://github.com/readium/ts-toolkit/tree/fxl-divina-improvements-backup

I prefer it published than leaving it locally and something bad happens to my machine, we do not have it any longer. 

I tried to bring Webpub to parity for `completeLocator` but it not having a positionsList with `totalProgression` it just treats reading order items as equal. In hindsight, not sure we should try to handle that instead of ignoring it.

PR #205, opened 2026-04-12
Add file extensions to all imports

smoores-dev:
The `@readium/*` npm packages declare themselves as `type: module`, but don't use file extensions in their generated type declarations (because the source files don't specify file extensions). This is incorrect — because ESM requires all non-bare import paths to specify a file extension, any typescript project with `modeResolution: "nodenext"` will simply be unable to use the type declarations. Typescript (silently, I might add) fails to resolve, e.g. `./fetcher` from `@readium/shared/types/src/index.d.ts`, because [using Node's ESM resolution rules](https://nodejs.org/api/esm.html#mandatory-file-extensions), that would require that there was a file named exactly `./fetcher`, but the actual file is named `./fetcher/index.d.ts`

 This PR just adds `.ts` extensions to every import in the codebase. Typescript now has an [`allowImportingTsExtensions` compiler flag](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) that allows you to, well, import `.ts` extensions. This works out fine for now because we're still bundling our dist files, but I will probably also open a PR to change that!
 
 Note for discussion: if we want to continue using vite/esbuild for transpilation, but stop bundling, we probably actually want to use `.js` extensions in our imports, not `.ts`, because esbuild doesn't support `rewriteRelativeImportSpecifiers`. If we instead switch to swc, then we can keep the `.ts` specifiers. Let me know what you're thinking on that front!


JayPanoz:
Thanks, this is greatly appreciated. 

I think merging this is a good first step towards rethinking distribution. It already solves an issue you have, and that way we can quickly align the ongoing PRs for audio, RTL and Divina, which are quite massive in features and improvements.  

For additional context, Vite was historically used as it was the easiest and surest way to handle Readium CSS import as `?inline` – it is now `?raw` with injection API. 

Then for distribution we went with the recommended Vite way at the time, to keep things simple. 

I think the RTL PR will be a good starting point to discuss options further, based on changes it brings to CSS and its imports since that was the major argument in favor of Vite 2 years ago. 

What's for sure is that we need to address build/bundle for Thorium Web anyway to reduce its build size so we have at least a second testing ground to see how things go.

smoores-dev:
That sounds good to me. Once everything has a file extension, it's also much easier to just find/replace in case we need to switch from .ts to .js to support whatever tooling we eventually land on. Or we could leave it and use [a plugin](https://github.com/sant123/esbuild-plugin-rewrite-relative-import-extensions) to do the rewrites if we stay with esbuild.

Also, on the note of unbundling — I think, in addition to (actually, contributing to) bundle size, there's an issue where the same package can end up included in the bundle multiple times, because each package is bundled separately, first, and dependencies are currently included in the bundle. This can mean that there are multiple instances of the same class definition, which don't get merged, which breaks some things (like instanceof checks).

Anyway, feel free to merge at any time, I think this is good to go, and feel free to reach out if you think I can help at all with unbundling! Happy to help

PR #209, opened 2026-04-28
Decorator API

JayPanoz:
This PR aims to implement the Decorator API per https://readium.org/architecture/proposals/008-decorator-api.html


JayPanoz:
cc @smoores-dev I know you have lots of experience with the API in Mobile so it will help tremendously if you could take a look. I tried filling in the gaps so that it stays aligned with Mobile toolkits but of course, you are way ahead on use cases, bugs, etc. 🙏

JayPanoz:
There will be more work happening on this but it should not impact the known public API, only the internals as we need to adjust for Readium Speech.

PR #218, opened 2026-05-15
fix: FXL last-page state not refreshing on revisit

mary-ext:
`positionChanged` was wired to only fire as a side effect of a page loading in, this fixes that so it always fires after every navigation.

encountered this when stresstesting Bookwalker's reader, we show a "last page reached" dialog upon reaching the last page, but it never gets shown again because Thorium's `atPublicationEnd` state became stuck, unless the user tried going back far enough that the last page gets evicted from the sliding window.

this possibly unblocks https://github.com/edrlab/thorium-web/issues/16.

### DIFF
--- a/navigator/src/epub/EpubNavigator.ts
+++ b/navigator/src/epub/EpubNavigator.ts
@@ -648,6 +648,11 @@ export class EpubNavigator extends VisualNavigator implements Configurable<Confi
                     }
                 }
             await this.apply();
+            // On the initial forward walk each fresh iframe load fires
+            // positionChanged via _pong; on back-then-forward to an
+            // already-loaded frame no _pong fires, so dispatch explicitly here.
+            // Redux dedupes the duplicate on the forward case.
+            this.listeners.positionChanged(this.currentLocation);
             return true;
         }
 


JayPanoz:
This is tied to the Thorium Web issue you mentioned indeed. Note it should have been resolved recently, as a side effect of another issue. 

I cannot recall the exact release but it was maybe a month or 2 ago, as the state handling for the publication start and end were missing in the FXL Mutation Observer. 

See also #66 for the TS toolkit specific tracking. 

My worry ever since we opened this issue has been the FXL resolution tainting reflowable – whose listener already triggers quite often on load, snapping, etc. – which is why I personally preferred a fix scoped to FXL, even if that was more work. 

That said it has also been stalling for lack of bandwidth, and if you did not notice a negative effect for reflowable we can probably go with this, knowing listeners is something we want to ideally redesign and rewrite this year anyways. 

I'll let @chocolatkey chime in just in case, and delegate the merging decision to him actually, since his ongoing PR touches EPUBNavigator and he may notice something I won't.

mary-ext:
this `positionChanged` call sits inside the `this._layout === Layout.fixed` branch, doesn't that address the tainting concern? unless the intent is for this FXL branch to be refactored, in which case fair enough and I'd close the PR if so.

I was aware Thorium had a fix already (which we haven't yet pulled), but I figured fixing the root cause was still preferable since the code around the MutationObserver workaround that Thorium is doing was marked with `[TMP]` comments.

JayPanoz:
Apologies, I was missing this detail in the GitHub app as contents were collapsed. That should address this concern.

JayPanoz:
Thanks again for the PR and your patience, I am only coming back to this after dealing with other tasks. 

I will merge this, release a new navigator package, remove the workaround in Thorium Web and prepare a patch release then, with all other opened PRs.

PR #220, opened 2026-05-23
fix: missing return in PublicationContentIterator

ddfreiling:
was trying to implement TTS features in web and happened upon this content iterator always returning null


chocolatkey:
Thanks for spotting this!

PR #235, opened 2026-06-24
Improve space media fragment

JayPanoz:
While fixing `LocatorLocations.time()` I noticed `space` was marked TODO as well. 

This improves parsing, although it is a breaking change as the Type was updated in the process.


JayPanoz:
@chocolatkey Note this is a breaking change in practice, due to the Return Type changing. It was not used in the codebase but it can impact existing consumers of shared models.

chocolatkey:
@JayPanoz thanks for letting me know. I actually just worked on this type of parsing for the Go toolkit, you can see it here if you want to compare the two implementations: https://github.com/readium/go-toolkit/blob/html-guided-navigation/pkg/guidednavigation/fragments.go#L449

PR #239, opened 2026-07-14
Fix blank iframe after remount by setting iframe.src on load

### LINKED ISSUE #238: FrameManager.load() never sets iframe.src, so browser remount/reload can blank the page
I have been using `@readium/navigator` and found a problem, the library can suddenly show a blank page (often after resize/layout changes) even though nothing in the app threw an error and the iframe is still there.

Tracing it, I notice that `FrameManager.load()` causes the iframe to navigate through `contentWindow.location.replace(source)` but never sets `iframe.src`.

After a successful load the document becomes visible, but the iframe’s `src` stays at its default (`about:blank` / empty). Any later browser navigation that goes through `src` (e.g. remounting/reparenting the iframe, or other `src`-based reload) restores `about:blank`. The iframe element remains in the DOM and looks fine; the app gets no load error from the navigator.

The same pattern seems to exist in `FXLFrameManager` and `WebPubFrameManager`.

```ts
// FrameManager.load()
this.frame.contentWindow!.location.replace(this.source);
// iframe.src is never set
```

So after `location.replace(blobUrl)` the document content is correctly loaded, but `iframe.src` is empty. So moving the iframe in the DOM (or otherwise reloading from `src`) then shows a blank page.

sandmor:
This fixes #238

The frame managers loaded blob content with `contentWindow.location.replace()` and never set `iframe.src`, so the attribute stayed at `about:blank`, provoking later for remount/reparent to restore a blank document while the iframe stayed in the DOM.

This fixes it via setting `iframe.src = source` in FrameManager, FXLFrameManager and WebPubFrameManager.

### DIFF
--- a/navigator/src/epub/frame/FrameManager.ts
+++ b/navigator/src/epub/frame/FrameManager.ts
@@ -66,7 +66,7 @@ export class FrameManager {
             this.frame.onerror = (err) => {
                 try { rej(err); } catch (error) {}
             }
-            this.frame.contentWindow!.location.replace(this.source);
+            this.frame.src = this.source;
         });
     }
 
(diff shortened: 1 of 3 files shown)


JayPanoz:
@chocolatkey I am unable to review this PR ar the moment because I do not have the all the background information relevant to `window.location.replace` and `iframe.src`. 

I know this was quite some time ago but would you happen to remember why you picked the former?

PR #256, opened 2026-09-21
Add optional config to prevent upgrading insecure connection

### LINKED ISSUE #252: Configurable `upgrade-insecure-requests` in CSP for `FrameBlobBuider`
Hey 👋 

I develop and maintain software that uses Readium and believe [this issue](https://github.com/stumpapp/stump/issues/1377) is a result of [`upgrade-insecure-requests`](https://github.com/readium/ts-toolkit/blob/develop/navigator/src/epub/frame/FrameBlobBuilder.ts#L9) being unconditionally added to the security policy on the frame. I made a [tentative fix](https://github.com/stumpapp/ts-toolkit/tree/optional-insecure-upgrading) on a fork that makes it conditional via a new config `upgradeInsecureRequests` that passes it down, but wanted to open an issue before assuming the change would be welcome in a PR. The default should preserve the current behavior (`always` upgrade).

This seems related-ish to https://github.com/readium/ts-toolkit/issues/120, though the referenced code snippet there is what I see on the develop branch so maybe that one is already resolved.

Thank you for all the work across all the Readium toolkits!

aaronleopold:
Closes #252

Happy to make any adjustments to get this in. It relates to a project I maintain: https://github.com/stumpapp/stump/issues/1377


JayPanoz:
@chocolatkey given this touches your previous work on securing iframes – and I'm sure you have other things in mind –, would you mind taking the lead for discussion/review on this one?

chocolatkey:
This looks like a good change, at the time I had forgotten about the case of non-https, non-localhost URLs. @JayPanoz do you foresee any issues with this? I would want to combine/merge this with my idea of more simpler "security levels" for the configuration, for less technical integrators but this needs a fix right now

JayPanoz:
No I'm completely fine as that is opt-in so not a breaking change. 

Will give it an extended look/review just in case I missed something yesterday.

